Issue #59, opened 2026-06-07
Line not updating when editing the geometry in the properties tab


nick-ziv:
After drawing a line, when I edit the coordinates manually in the properties panel, it updates the handle positions but does not redraw the line itself.  See the images below for before and after editing the start X of a line:
Before:
<img width="860" height="360" alt="Image" src="https://github.com/user-attachments/assets/8c17bdaf-7e0f-49bf-a2e4-52a5fc77a158" />
After (edited then pressed enter):
<img width="860" height="360" alt="Image" src="https://github.com/user-attachments/assets/3df48a57-6327-4107-b973-94cf359ccabd" />

HakanSeven12:
fixed

PR #66, opened 2026-06-07
Implement Expression Evaluation for numerical inputs and the CLI

nick-ziv:
This PR implements an expression evaluation feature to inputs across the UI.  This includes the properties manager, the CLI, and the dynamic overlay fields.  This is a feature present in most commercial CAD software. 

This implementation enables the following:

## Operators

| Operator | Meaning | Example | Result |
|---|---|---|---|
| `+` | Addition | `10+20` | `30` |
| `-` | Subtraction | `10-3` | `7` |
| `*` | Multiplication | `10*5` | `50` |
| `/` | Division | `100/4` | `25` |
| `^` | Power | `2^3` | `8` |
| `%` | Modulo | `10%3` | `1` |

## Functions

| Function | Description | Example | Result |
|---|---|---|---|
| `abs(x)` | Absolute value | `abs(-5)` | `5` |
| `sin(x)` | Sine (radians) | `sin(0)` | `0` |
| `cos(x)` | Cosine (radians) | `cos(0)` | `1` |
| `tan(x)` | Tangent (radians) | `tan(0)` | `0` |
| `sqrt(x)` | Square root | `sqrt(4)` | `2` |
| `ceil(x)` | Round up | `ceil(3.1)` | `4` |
| `floor(x)` | Round down | `floor(3.9)` | `3` |
| `round(x)` | Round nearest | `round(3.7)` | `4` |
| `log(x)` | Natural log | `log(1)` | `0` |
| `exp(x)` | e^x | `exp(0)` | `1` |

## Coordinate Functions

| Function | Description | Example | Result |
|---|---|---|---|
| `PtDist(x1,y1,x2,y2)` | Distance between two points | `PtDist(0,0,3,4)` | `5` |
| `PtAngle(x1,y1,x2,y2)` | Angle from point 1 to point 2 (radians) | `PtAngle(0,0,3,4)` | `0.927...` |

## Constants

| Constant | Value |
|---|---|
| `pi` | 3.14159... |

Screenshot:
Screenshot from 2026-06-07 13-10-18
turns into...
Screenshot from 2026-06-07 13-10-33

Known issues:
- This is existing but is very obvious when testing: Issue #59 
- The dynamic fields near the crosshair do not allow moving the cursor using arrow keys
- The dynamic fields do not support typing in letters so the named functions cannot be used

### DIFF
--- a/src/app/expr_eval.rs
+++ b/src/app/expr_eval.rs
@@ -1,0 +1,598 @@
+/// Math expression evaluator — recursive-descent parser.
+///
+/// Supports operators: `+ - * / ^ %` with standard precedence.
+/// Supports functions: `abs sin cos tan sqrt ceil floor round log exp`
+/// Supports coord functions: `PtDist PtAngle`
+/// Supports constant: `pi`
+/// Case-insensitive for function names and constant.
+
+// ── Token ───────────────────────────────────────────────────────────
+
+#[derive(Debug, Clone, PartialEq)]
+enum Token {
+    Number(f64),
+    Ident(String),
+    Plus,
+    Minus,
+    Star,
+    Slash,
+    Caret,
+    Percent,
+    LParen,
+    RParen,
+    Comma,
+}
+
+// ── Tokenizer ───────────────────────────────────────────────────────
+
+fn tokenize(input: &str) -> Result<Vec<Token>, ()> {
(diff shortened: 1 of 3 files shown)


HakanSeven12:
I was thinking of adding this feature as well.

PR #105, opened 2026-06-15
Temporarily suppress Ortho mode during RECTANG commands

KarimJerbi:
When a rectangle command (RECT, RECTANG, REC, RECT_ROT, RECT_CEN) is started with Ortho mode on, ortho is automatically turned off for the duration of the command and restored when it finishes.

If the user manually toggles ortho (F8) during the command, that toggle is treated as permanent and no auto-restore occurs on exit.

Changes:

- Added rect_suppressed_ortho: bool flag to track when a rectangle command suppressed ortho
- Set the flag and disable ortho on command start (all 3 rectangle entry points)
- Restore ortho on command exit via restore_pre_cmd_tangent
- Clear the flag on manual ToggleOrtho so user toggles are permanent

### DIFF
--- a/src/app/cmd_result.rs
+++ b/src/app/cmd_result.rs
@@ -1674,12 +1674,17 @@ impl OpenCADStudio {
         self.focus_cmd_input()
     }
 
-    /// Restore the tangent snap state that was in effect before the command started.
+    /// Restore the tangent-snap / ortho state that was in effect before the command started.
     fn restore_pre_cmd_tangent(&mut self) {
         if let Some(was_on) = self.pre_cmd_tangent.take() {
             if !was_on {
                 self.snapper.enabled.remove(&crate::snap::SnapType::Tangent);
             }
+        }
+        if self.rect_suppressed_ortho {
+            self.rect_suppressed_ortho = false;
+            self.ortho_mode = true;
+            self.polar_mode = false;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


HakanSeven12:
Thanks for this — the Ortho-suppression feature itself looks good: the `rect_suppressed_ortho` flag, disabling on the three rectangle entry points, restoring in `restore_pre_cmd_tangent`, and treating a manual F8 during the command as permanent is a clean design.

However, the PR can't be merged as-is because of an **unrelated regression in `src/app/update.rs`**. The branch was cut from `e31dbcb`, and the commit reverts recent dynamic-input work that lives in that file. Compared to current `main`, this PR removes:

- the `dyn_ref` usage (anchor reference for the perpendicular guide) — 5 uses → 0
- the Width/Height `copysign(live)` signed-entry logic in the `val` closure
- the `Perp` branch in `dyn_resolve_point`
- the `dyn_ref_screen` projection in the cursor-move handler

GitHub reports the merge as clean, so this would silently regress: **RECT_ROT height (perpendicular drop), ELLIPSE minor-axis Perp measure, and Width/Height signed input**. The branch is also missing the newer `main` commits (`5128d94`, `9b07edb`).

Could you please:
1. rebase onto current `main`,
2. drop the `update.rs` dynamic-input reversions (keep only the `rect_suppressed_ortho = false` line in the `ToggleOrtho` handler),
3. keep just the Ortho hunks in `cmd_result.rs` / `commands.rs` / `mod.rs`.

That should leave the diff to only the Ortho-suppression change. Happy to help if the rebase is fiddly.

KarimJerbi:
i think this happened because i run git add -A instead of just git add, should be solved now !

HakanSeven12:
thank you

PR #127, opened 2026-06-17
Add TORIENT command

KarimJerbi:
This PR implements the TORIENT command, which allows users to rapidly orient selected text objects to a specific angle or make them "Most Readable" without changing their physical positions in the drawing, I use this command frequently.

- Supported Entities: Works with Text, MText, AttributeDefinition (ATTDEF), and Insert (BLOCK) references.
In-Place Attribute Rotation: For Insert blocks, it iterates through and intelligently rotates the attached Attribute objects without rotating the parent block itself.
- "Most Readable" Default: Enter applies the "Most Readable" logic, which automatically rotates upside-down text so it reads correctly relative to the viewport.
- Absolute Orientation: specify a new absolute angle by typing a numerical value or drawing a 2-point vector on the canvas.
- Selection Support: supports both pre-selection (selecting objects then running the command) and post-selection

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -141,6 +141,7 @@ Status of every standard CAD command in Open CAD Studio:
 | `FIELD` | — | Auto-updating text field | ❌ |
 | `SPELL` | SP | Spell check | ❌ |
 | `ARCTEXT` | — | Text along an arc | ❌ |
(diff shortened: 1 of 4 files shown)


HakanSeven12:
thank you

PR #130, opened 2026-06-18
feat: implement TEXTEDIT command and TEXTEDITMODE system variable

KarimJerbi:
- Adds TEXTEDIT command to seamlessly edit text objects in-place.
- Matches industry-standard CAD tools' prompt behavior and "Current settings" banner logic.
- Implements the TEXTEDITMODE system variable (0/1) for Multiple or Single editing modes.
- command lifecycle with CmdResult::SuspendForTextEdit and post_editor_closed hooks, allowing commands to cleanly pause and resume after external UI editors close.

### DIFF
--- a/src/app/cmd_result.rs
+++ b/src/app/cmd_result.rs
@@ -1617,6 +1617,46 @@ impl OpenCADStudio {
                 self.ribbon.deactivate_tool();
                 return self.begin_text_edit(handle);
             }
+            CmdResult::SuspendForTextEdit { handle } => {
+                let is_editable = self.tabs[i]
+                    .scene
+                    .document
+                    .get_entity(handle)
+                    .map(|e| crate::app::text_inline::read_text_field(e).is_some())
+                    .unwrap_or(false);
+                if !is_editable {
+                    self.command_line.push_error("TEXTEDIT: selected entity is not text.");
+                    let prompt = self.tabs[i].active_cmd.as_ref().map(|c| c.prompt());
+                    if let Some(p) = prompt {
+                        self.command_line.push_info(&p);
+                    }
+                    return Task::none();
+                }
+                let cmd = self.tabs[i].active_cmd.take();
+                self.tabs[i].suspended_cmd = cmd;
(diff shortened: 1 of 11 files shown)


HakanSeven12:
## Review

Solid design — the suspend/resume command lifecycle is clean and the guards are in the right places. A few findings:

### 🐛 Bug — invalid `TEXTEDITMODE` value aborts the command
In `TexteditmodeCommand::on_text_input` (`src/modules/annotate/textedit.rs`), an unrecognized value returns:
```rust
Some(CmdResult::Measurement("Requires 0 OR 1 OR MULTIPLE OR SINGLE".to_string()))
```
But the `Measurement` arm in `src/app/cmd_result.rs:707` sets `active_cmd = None`, which **ends the command**. So a typo at the value prompt cancels `TEXTEDITMODE` instead of re-prompting. Expected: print the error and keep asking.

### ♻️ Duplication — `0/1/m/s/multiple/single` parsed in three places
The same value parsing lives in:
- `src/app/commands.rs:4923` (direct `TEXTEDITMODE <val>` entry)
- `TexteditmodeCommand::on_text_input`
- `src/app/settings.rs:145` (settings load)

Worth extracting a single `parse_texteditmode(&str) -> Option<bool>` helper.

### 🧹 Trailing whitespace
`src/modules/annotate/textedit.rs` has trailing whitespace on lines 100, 113, 121, 130, 201, 208.

### ✅ Looks good
- `post_editor_closed`'s `None`-guard keeps the normal MTEXT/TEXT flow safe.
- `can_edit_text`'s Leader resolution loop is bounded (no cycles).
- `texteditmode` persists automatically via `current_settings`.
- Prompt redisplay works through the `NeedPoint` arm, so Mode / Single-Multiple switching shows the right prompt.

No blockers. The `Measurement` abort is the one worth fixing before merge.

KarimJerbi:
do i need to adress the duplications and mesearment arm ? sorry i took my time

HakanSeven12:
No 🙂 I fixed it myself

Issue #145, opened 2026-06-20
Protect host from plugins


schoeller:
Dear all,

thanks for the work.

In #134 a maldesigned plugin crashed the host by chance. This should never be possible. Having mistral + leanstral investigate it recommends to alter the loading architecture as follows: 

> ```
> **Core Problem:** The current architecture loads plugins as `cdylib` libraries in the **same process** as the host. This means:
> - Plugins share the host's memory space
> - Plugins can crash the host via panics or undefined behavior
> - The host cannot protect against plugins using `unsafe` to violate Rust's safety guarantees
> - `catch_unwind` cannot catch Undefined Behavior (UB) from `transmute_copy`, dangling pointers, etc.
> 
> **Solution:** Load plugins in **separate processes** with **Inter-Process Communication (IPC)**. This provides **true isolation** - a plugin crash cannot affect the host process.

If this is the correct design approach is not up to my competence.

HakanSeven12:
I put a guard between host and plugin layers. Separate possess  drops the performance so pass for now

schoeller:
Stil crashing on W11 v0.6.3. 

1. Download https://github.com/user-attachments/files/29124292/opencad.pyshell.zip
2. Extract and copy folder `opencad.pyshell` into `%APPDATA%\Roaming\OpenCADStudio\plugins`
3. Restart OCS and open new document
4. Click Python ribbon + Python => Crash

HakanSeven12:
If you install github cli and give access to AI you can manage your code base over github. So you dont need to download upload your code everytime 🙂

schoeller:
The workflow I have: Origin -> Fork -> Pull -> VS Code + AI -> Testing -> Push -> Pull request to origin

I am using currently Granite 4.1 for local analysis and prep of phasing. Then VS Code + Mistral to implement. Granite 8.1 4B is running on local 8BG VRAM ok, but the larger jobs kill it somehow. Dunno why yet. It's my first steps. Difficult, cause I cannot review myself which is dangerous.

HakanSeven12:
If you have this code in a repo can you give me the repo url?

schoeller:
> If you have this code in a repo can you give me the repo url?

[...]

PR #172, opened 2026-06-24
Protect host from plugins: efforts for review

schoeller:
Dear all,

by coincide the host was crashed with a plugin, see https://github.com/HakanSeven12/OpenCADStudio/issues/145. This pull request tries to seperate memory and processes between host and plugin in order to protect the host.
It tries to achieve the following targets: 0) backwards compatibility, 1) process separation, 2) memory separation, 3) performance issues, 4) minimal change in `src/` and 5) concentrating changes in `crates/ocs_plugin_api` with 6) a new template in `docs/plugin-template-v2`.

It has been rudimentally tested.

Best

Seb


schoeller:
Fixed plugin_interactive_command_drives_host_flow to double precision.

HakanSeven12:
Thanks @schoeller — this is a substantial piece of work and it's now on `main`. I rebased the 7 commits onto current `main` (the branch had fallen ~50 commits behind, including the f64 pick-coordinate migration) and added two follow-up fix commits on top:

**`fix(plugin)`** (b409372)
- Converted the out-of-process interactive adapter's `on_point`/`on_entity_pick` to `glam::DVec3` — the `CadCommand` trait moved to f64 pick coordinates after this branch was cut, so it no longer compiled against the trait.
- Contained plugin-constructor panics in `export_plugin!`: `ocs_plugin_register` runs across the C ABI boundary, where an unwinding panic is UB. It now catches and returns null (the loader already treats that as a failed registration).
- Reaped killed runner processes: the spawn-timeout and disconnect paths called `child.kill()` without `wait()`, leaving a zombie on Unix. Both paths and `shutdown()` now route through a shared `reap()` (kill + detached wait).

**`fix(xclip)`** (deec2a8) — unrelated drive-by exposed once tests compiled again.

The process-isolation design (length-framed IPC + 64 MiB cap, `catch_unwind` around every plugin call in the runner, double-buffered shared memory with rkyv validation) is solid and does what #145 needed.

One known follow-up left for a separate change: there's no read timeout on the IPC stream, so a plugin that *hangs* (rather than crashes — crashes are already handled via broken-pipe → cancel) can still block the host UI thread. `interprocess::Stream` has no portable timeout API, so it needs a per-call watchdog; out of scope here.

Landed via rebase (different SHAs), so GitHub won't auto-close — closing manually. Your work is on `main`.

schoeller:
> One known follow-up left for a separate change: there's no read timeout on the IPC stream, so a plugin that _hangs_ (rather than crashes — crashes are already handled via broken-pipe → cancel) can still block the host UI thread. `interprocess::Stream` has no portable timeout API, so it needs a per-call watchdog; out of scope here.

@HakanSeven12 You seem to never sleep. I understand the follow-up `timeout` is on the task list. Whenever plugin-designers have accepted the design change I will investigate further.

HakanSeven12:
> You seem to never sleep. 

sometimes :)

schoeller:
BTW and for the record. I asked Kimi 2.7 to analyzed MAIN in some fashion. It has nothing to do with stabilizing functionality for day-to-day use.

This is what it came up with:

# OpenCADStudio Codebase Review

Scope: `src/` and `crates/ocs_plugin_api` in `OpenCADStudio-main`.

## 1. Current Security

| Area | Observation |
|------|-------------|
| Network | `update_check.rs` and `plugin/marketplace.rs` fetch JSON/assets over HTTPS via `ureq`+`rustls`. No certificate pinning, no checksum/signature verification, and user-supplied `owner/repo` strings are interpolated directly into URLs. |
| Plugin runtime | Plugins are loaded **out-of-process** (`PluginProcess`/`runner.rs`) — good isolation. However, the cdylib is still `dlopen`-ed inside the runner with `libloading` (`runner.rs:148`) and the runner executable is spawned from the host binary with no code-signature check. |
| Plugin install | `marketplace::install` downloads arbitrary GitHub release assets and writes them into the user plugins directory. No sandbox, no signature verification, no permission model. |
| Plugin manifest | `plugin/external.rs` parses `plugin.toml` with a hand-rolled line parser; `OCS_PLUGINS_DIR` lets any directory be treated as the plugin store. |
| Automation server | `--serve` exposes a JSON line protocol on `stdin/stdout` or `127.0.0.1:<port>`. No authentication, no capability ACL; `open`, `save`, `run` operate on caller-supplied paths. |
| File associations | `io/file_association.rs` writes Windows registry keys, Linux `.desktop` files, and calls macOS LaunchServices. These are privileged OS surfaces and the code uses `unsafe` COM/CoreFoundation bindings. |
| Memory safety | ~20 `unsafe` blocks: `libloading` in runner, `memmap2` in `shm.rs`, Windows COM in `file_association.rs`, printer APIs. `shm.rs` uses atomics for double-buffering; `check_archived_root` is used, which is safer than unchecked deserialization. |
| Error handling | 125 `.unwrap()`, 38 `.expect()`, 15 `panic!` (source scan). Several are in plugin/IPC/shm paths where a malicious or malformed plugin could trigger a host panic. |
| Input parsing | Hand-rolled JSON string extractor in `update_check.rs`, hand-rolled TOML parser, DWG/DXF parsing via `acadrust`. No obvious input-length caps in the update-check parser. |
| Clipboard / Web | Web clipboard read is gated on user gesture; OK. `sys.rs` percent-encodes query data correctly. |

## 2. Security Optimizations (Platform-Independent)

1. **Cryptographic verification for plugins & marketplace**
   - Sign every plugin release with a project key (e.g., `minisign`/`ed25519`).
   - Ship a pinned set of trusted public keys; reject any plugin whose signature does not verify.
   - Verify SHA-256 checksums of downloaded assets before writing them to disk.

2. **Least-privilege plugin sandbox**
   - Define a capability manifest (`files`, `network`, `clipboard`, `printer`) in `plugin.toml`.
   - Run plugin processes with no filesystem access except a dedicated workspace directory and the shared-memory document view.
   - Reject plugin requests that exceed declared capabilities.

3. **Automation hardening**
   - Require a session token for TCP socket mode; bind only to `127.0.0.1` by default.
   - Validate all paths against an allow-list and resolve them with `std::fs::canonicalize` before use.
   - Add an opt-in read-only mode that refuses `save`/`run` commands.

4. **Parser hardening**
   - Replace the hand-rolled `plugin.toml` parser with `serde` + `toml` and schema validation.
   - Replace the hand-rolled GitHub-JSON extractor with `serde_json`.
   - Add length limits, recursion limits, and fuzz tests for DWG/DXF/SHM deserialization.

5. **Eliminate panic paths in security-critical code**
   - Audit and remove `unwrap`/`expect`/`panic!` in plugin host, IPC, SHM, and automation.
   - Propagate errors to the UI as `Result` instead of crashing.

6. **Secure defaults**
   - Disable the marketplace and automation server unless explicitly enabled.
   - Pin TLS roots and enable certificate transparency where possible.
   - Zero shared-memory files before unlinking.

## 3. Design Optimization

1. **Break up the god object**
   - `OpenCADStudio` (`app/mod.rs`) holds hundreds of fields. Split it into focused controllers: `DocumentController`, `CommandController`, `PluginController`, `ViewportController`, `UIState`.

2. **Decouple UI from domain logic**
   - Route all user actions through `StepInput`/`CmdResult` abstractions (already partially done). Ensure `update.rs` does not directly mutate document internals; go through the command controller.

3. **Unify document snapshot paths**
   - Today the document is serialized for save/load, XDATA round-trips, and SHM views through separate code. Introduce a single `DocumentSnapshot` type and reuse it.

4. **Typed dialog/state buffers**
   - Dozens of `String` edit buffers (`ds_dimasz`, `page_setup_w`, etc.) store raw user input. Replace them with validated value types (`Length`, `Angle`, `Scale`) that parse on edit and report errors centrally.

5. **Plugin manifest schema**
   - Use `serde`/`toml` and a struct `PluginManifest` instead of the hand-rolled parser. This removes duplicate parsing logic and enables better error messages.

6. **Command registration**
   - The `inventory`-based registry is fine, but command metadata (name, aliases, required capabilities) should be declarative and documented.

## 4. Architecture Optimization

1. **Layered architecture**
   - UI (`iced` widgets) → Application Services → Domain (`acadrust`/`CadDocument`) → Infrastructure (file IO, GPU, plugins, network).
   - Depend on traits/abstractions, not concrete `OpenCADStudio`.

2. **Plugin API versioning**
   - Stabilize `HostApi` as a versioned C ABI or a capability-based IPC contract.
   - Keep `ocs_plugin_api` small and `#![deny(missing_docs)]`; do not expose internal `acadrust` types directly unless intended.

3. **Repository pattern for persistence**
   - Abstract `settings`, recent-files, plugin registry, and update cache behind a `SettingsRepository` trait. The concrete implementation can still be platform-specific, but the interface is portable.

4. **Separate headless binary**
   - Move `--serve` and `--export` into a `ocs_automation` binary that links the core library. This removes CLI/automation concerns from the GUI main.

5. **Async I/O**
   - Replace blocking `ureq` calls with an async HTTP client integrated with `iced::Task`. Add timeouts, retries, and cancellation.

6. **Error architecture**
   - Standardize on `thiserror` for library errors and `anyhow` for application glue. Never panic across the plugin boundary.

7. **Rendering data model**
   - `Scene` currently mixes domain state (`document`), GPU caches, and viewport state. Split into `Scene` (domain) and `RenderCache` (GPU-derived data), with cache invalidation driven by a small event log.

## 5. Documentation Optimization

1. **Enforce rustdoc**
   - Add `#![warn(missing_docs)]` to `src/lib.rs` and `#![deny(missing_docs)]` to `crates/ocs_plugin_api`.
   - Document every public struct, trait, enum variant, and method.

2. **Architecture Decision Records**
   - Create `docs/adr/` entries for: out-of-process plugins, shared-memory document views, world-offset rendering, and the command `StepInput`/`CmdResult` design.

3. **Threat model & security docs**
   - Add `docs/security.md` describing the plugin trust model, marketplace verification, and automation server risks.

4. **Plugin author docs**
   - Extend `docs/plugin-architecture.md` with a complete `HostApi` reference and example plugin.

5. **README restructuring**
   - Keep README as a quick-start. Move build details to `docs/build.md`, packaging to `docs/packaging.md`, and command reference to `docs/commands.md`.

6. **Intra-doc links**
   - Replace free-form references like "see `app::update`" with rustdoc `[`Link`]` syntax.

HakanSeven12:
create an issue please

PR #173, opened 2026-06-24
fix(save-dialog): allow navigating to other drives (#170)

### LINKED ISSUE #170: save to.. able to select other drives
when I "save as" a file on a particular drive or usb stick I need to save it to a different drive
at the moment it seems this is not possible
thanks

KevinGriffin-new:
Fixes #170.

## Problem
The custom Save As dialog could only navigate up the tree via `Path::parent()`, which returns `None` at a drive root (`C:\`). Navigation therefore dead-ended on the starting volume — other drives and USB sticks were unreachable, so a drawing could not be saved to a different drive.

## Fix
Add a synthetic Windows **"This PC"** drives root above the drive letters:

- `parent_folder()` now climbs `…\sub → C:\ → This PC → None`.
- `read_dir_entries()` lists the available volumes when at that root.
- The dialog's up-button (`↑`), the `..` entry, and the path bar all route through the same helper; the path bar shows "This PC" at the top.

Volumes are probed `A:`–`Z:` by stat — **no new dependency** and no Win32 binding. All Windows-specific behavior is behind `#[cfg(windows)]`, so non-Windows builds are unchanged.

Resulting navigation: `C:\Users\me → C:\Users → C:\ → This PC → D:\` (or a USB stick).

## Notes
- Removable drives with **no media inserted** are skipped (the stat fails) until media is present — they can't be entered anyway.
- Same dialog backs plain Save on an unsaved doc and the unsaved-changes prompts, so those inherit the fix.

## Testing
- `cargo check` passes with no new warnings.
- On a two-drive machine the volume probe correctly lists `C:\` and `D:\`.

### DIFF
--- a/src/app/view.rs
+++ b/src/app/view.rs
@@ -3264,8 +3264,12 @@ fn save_as_dialog_window<'a>(
     };
 
     // ── Path bar ─────────────────────────────────────────────────────────
-    let path_str = folder.to_string_lossy().into_owned();
-    let up_path = folder.parent().map(|p| p.to_path_buf());
+    let path_str = if crate::io::is_drives_root(folder) {
+        "This PC".to_string()
+    } else {
+        folder.to_string_lossy().into_owned()
+    };
+    let up_path = crate::io::parent_folder(folder);
     let path_bar = row![
         {
             let up_msg = up_path.map(Message::SaveDialogNavigate);
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -182,14 +182,72 @@ fn resolve_image_file(raw: &str, base_dir: Option<&Path>) -> Option<String> {
 
 // ── Directory listing (used by the custom Save As dialog) ────────────────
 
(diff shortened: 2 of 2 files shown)


HakanSeven12:
Thanks @KevinGriffin-new! Merged this into `main` (commit d161311) and extended the same volumes-root mechanism to macOS and Linux on top of it (125b89d):

- `parent_folder()` now climbs the filesystem root into the volumes list on every platform (`/` → Computer), not just Windows.
- `available_drives()` is platform-specific: Windows probes `A:`–`Z:`, macOS lists `/` + `/Volumes/*`, Linux lists `/` + the removable mounts under `/run/media/$USER`, `/media/$USER` and `/mnt`. Other Unix and wasm fall back to `/`.
- The path bar shows "This PC" on Windows and "Computer" elsewhere.

Since it landed via a rebase (different SHA), GitHub won't auto-close this. Closing manually — your fix is on `main`.

Issue #174, opened 2026-06-25
Apple Notarize the OSX release


KevinGriffin-new:
I had to develop a small "sign Mac OS dmgs" skill for one one of my projects: https://github.com/KevinGriffin-new/music-chop/blob/main/.claude/skills/sign-macos/SKILL.md

If we wanted to do something similar for OpenCADStudio:
either:
HakanSeven12 or the foundation gets Apple Developer ID. I'm contributing a version of the skill recipe in my fork
Or
  I sign with my Developer ID → works mechanically (the skill auto-detects whatever identity is present), but then every official OCS release is signed "KevinGriffin-new" and my Apple app-specific password lives in the OpenCADStudio  repo secrets.
 That's really the maintainer's call

HakanSeven12:
The **Open EAC** foundation will provide the Apple Developer ID; releases get signed/notarized under its identity, with the secrets and `release.yml` wiring done in-house. Closing — identity question settled.

PR #175, opened 2026-06-25
Add sign-macos skill (Developer ID sign + notarize recipe)

KevinGriffin-new:
Groundwork for #174 (Apple-notarize the macOS release).

Adds a project-scoped Claude skill at `.claude/skills/sign-macos/SKILL.md` — the
end-to-end recipe for a Gatekeeper-clean, signed + notarized macOS `.app`/`.dmg`
(Developer ID, outside the App Store), mapped onto OpenCADStudio's `build-macos`
job. It's a generalized version of one I wrote for another project.

**Scope: docs/skill only — this does NOT change `release.yml` yet.** The actual
workflow wiring (replace the ad-hoc signature with a Developer-ID sign + notarize
+ staple, gated on repo secrets with an ad-hoc fallback for forks) is a follow-up
once the **"whose Developer ID"** decision in #174 is settled.

Notes specific to OCS, captured in the skill's footer:
- It's a single Rust Mach-O, so **no entitlements are needed** — hardened runtime
  alone notarizes (the JIT/unsigned-memory entitlements are Python/V8 things).
- The five Apple secrets the workflow would need (`MACOS_CERT_P12`,
  `MACOS_CERT_PASSWORD`, `APPLE_ID`, `APPLE_APP_PASSWORD`, `APPLE_TEAM_ID`) —
  mirroring how the Windows job already pulls `AZURE_*`.
- The identity is auto-detected from the imported cert, so nothing is hardcoded.

Happy to follow up with the `release.yml` change whenever the cert question lands.

### DIFF
--- a/.claude/skills/sign-macos/SKILL.md
+++ b/.claude/skills/sign-macos/SKILL.md
@@ -1,0 +1,158 @@
+---
+name: sign-macos
+description: Sign and notarize a macOS app bundle (.app) + its .dmg for Developer ID distribution outside the Mac App Store. Use when producing a distributable signed/notarized Mac build, or when an `xcrun notarytool` submission comes back `Invalid` and needs diagnosis + a re-sign. Auto-detects the Developer ID identity from the keychain, so no Team ID is hardcoded. The recipe is project-agnostic; an "Applying it here" footer maps it onto this repo.
+---
+
+# Sign + notarize a macOS app (Developer ID)
+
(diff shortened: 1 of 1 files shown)


HakanSeven12:
The **Open EAC** foundation will provide the Apple Developer ID and we'll wire up signing in-house (see #174), so we won't vendor an external skill. Closing as invalid — thanks for the input.

PR #194, opened 2026-06-26
fix(dist): measure in f64 so snapped endpoints stay precise (#193)

### LINKED ISSUE #193: DIST with snap endpoints not very accurate
Command DIST, using Snap endpoint to endpoint is getting values near the Endpoint (mouse position?), instead of real endpoint position.

KevinGriffin-new:
## Problem (#193)
`DIST` reports values slightly off the real snapped endpoint. The snap is correct — the loss is in the command. `DistCommand` stored the first point as `Vec3` and `on_point` began with `let pt = pt.as_vec3();`, downcasting the f64 picked point (`DVec3`) to f32 before any math. At survey-scale coordinates (eastings ~5e5) f32's 24-bit mantissa only resolves to ~0.03–0.06 units, so the measurement lands *near* the endpoint rather than on it.

## Fix
Store `first` as `DVec3` and keep the distance/delta/angle math in f64; cast to f32 only for the visual preview wire vertices.

Verified with `cargo check --lib`.

Closes #193.

### DIFF
--- a/src/modules/draw/inquiry/dist.rs
+++ b/src/modules/draw/inquiry/dist.rs
@@ -1,12 +1,15 @@
 // DIST command — measure distance and angle between two picked points.
 
-use glam::{DVec3, Vec3};
+use glam::DVec3;
 
 use crate::command::{CadCommand, CmdResult};
 use crate::scene::model::wire_model::WireModel;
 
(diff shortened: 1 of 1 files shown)


HakanSeven12:
Thank you

PR #197, opened 2026-06-26
Swapping hand-rolled toml-parser againt serde+toml

schoeller:


HakanSeven12:
Reviewed and verified locally: merges cleanly into current `main` and `cargo build` is clean. The `Result<_, String>` signature (logged in `discover()` instead of a silent skip) and dropping the hand-rolled parser for standard TOML are both nice improvements. The real `landsurvey` `plugin.toml` (proper `[plugin]`/`[opencad]` sections) parses fine.

A few small notes, none blocking:

1. **Schema is now section-strict.** The old parser was section-agnostic (it matched keys regardless of which `[table]` they sat under), so a flat or mis-sectioned `plugin.toml` would still parse. With serde, `id` must be under `[plugin]` and the rest under `[opencad]`. That's a correctness improvement, but it's a behavioral tightening worth calling out — any previously-tolerated malformed manifest will now error (with a clear message, which is the upside).

2. **`xdata_apps` is captured but never read.** The new field is added to `ExternalPlugin` and parsed, but nothing in the host consumes `ExternalPlugin.xdata_apps`. Fine if it's staged for future use — just flagging it so it isn't mistaken for wired-up behavior.

3. **Nit:** the `Cargo.toml` comment says "serde is also used by marketplace JSON" — that path uses `serde_json`; `serde` (with `derive`) wasn't a direct dependency before this PR. Trivial wording only.

LGTM overall.

schoeller:
It probably should not be too tight. So maybe better to step back and drop it. Too much mingling overdoes it (also with [Shared-memory DocumentReader](https://github.com/HakanSeven12/OpenCADStudio/blob/a5023790481a7985e3def1a3706cc321943c327d/docs/plugin-template-v2/DESIGN.md) )

PR #208, opened 2026-06-28
Adding runner handshake

schoeller:
**Plugin IPC impersonation** — any same-user process can connect to the predictable local-socket name and pretend to be the host or runner.

- **Runner authentication model:** Pre-shared token delivered via the child process environment (`OCS_PLUGIN_TOKEN`). The runner presents the token in its first IPC frame; the host drops connections that do not.
  - Rationale: Cross-platform, minimal changes to the existing spawn/connect flow, no extra OS pipes or ACL code.
  - Limitation: A same-user process that can read the child’s environment can steal the token. This is accepted as a known residual risk.

- Out-of-process plugins authenticate with an env-token and run as separate processes, but are otherwise fully trusted.


HakanSeven12 (approved):
Token auth sound; handshake read now bounded by a deadline (fixup 7e54d28). LGTM.

PR #209, opened 2026-06-28
Adding expression evaluator against stack-overflow DoS

schoeller:


HakanSeven12:
Merged into `main` via cherry-pick as bfb81f6 (kept your authorship). Thanks!

Follow-up 87232bb bounds the production entry point `eval_to_string` against a separate super-linear (≈O(n³)) wall-clock DoS the depth guard didn't cover: deeply-nested / long non-evaluable input reached `eval_to_string` (command line, viewport, dynamic input — synchronous on the UI thread) and froze the app for seconds-to-minutes (measured 4 KB of parens → 51 s) even though `eval_number` fast-returns None. Capped the embedded-expression fallback scan; whole-string expressions of any length still evaluate via the fast paths.

Issue #177, opened 2026-06-25
some text font not showing


steverugi:
please see drawing [DWG](https://www.dropbox.com/scl/fi/vf22tmb35ip7faw7sstlt/2026-06-25-Prova.dwg?rlkey=rirx3dtrgtw22j1quy7owdiep&st=vrx3dq1d&dl=0) here

when opened in QCAD or DWG TrueView text shows correctly, not the same in OCS, (external images do show fine, links work well) 

I then changed the font in QCAD (I couldn't find a way directly in OCS) and reopened it and all was showing ok, maybe use of an alternative font available would sort it out?

Thanks

## QCAD

<img width="870" height="596" alt="Image" src="https://github.com/user-attachments/assets/6a8a93d9-af48-48cc-8d1a-3e1061b1b782" />

## TrueView

<img width="857" height="613" alt="Image" src="https://github.com/user-attachments/assets/ba5e6a91-8725-4355-9a04-c8144bb6d0ad" />

## OCS

<img width="1053" height="765" alt="Image" src="https://github.com/user-attachments/assets/d39fd523-c1d2-4d59-8062-4b80688a6906" />

KevinGriffin-new:
# Comment for #177 — "some text font not showing"

Root cause looks like **inline MTEXT font overrides (`\f...`) not being honoured**.

Dumping the attached `Prova.dwg`, the text that disappears in OCS is MTEXT whose
content carries an inline font switch, e.g.:

```
MTEXT  style='StileQuote'  value="{\fCentury Gothic|b0|i0|c0|p34; ... }"
MTEXT  style='Standard'    value="{\fCentury Gothic|b1|i0|c0|p34; CASH DESK ... }"
```

The text *style* (`StileQuote`) resolves to `arial.ttf`, but the MTEXT body
overrides the face to **Century Gothic** via the `\f` formatting code. QCAD and
DWG TrueView apply the `\f` override (or fall back to a default face) and render
the text; OCS renders these runs blank. That also matches the reporter's fix —
changing the font in QCAD (which drops/rewrites the `\f` run) makes it show in
OCS.

So the gap is in the MTEXT formatting-code handling: when an MTEXT run selects a
font via `\f<FontName>` that OCS can't resolve, it should **fall back to the
style's font (or a default)** rather than drawing nothing — and more generally,
parse and honour the `\f` face override the way the other readers do.

(External images and links in the same drawing render fine — this is specifically
the MTEXT `\f` inline-font path.)

[...]

PR #211, opened 2026-06-28
fix(block-cache): per-wire colour, fill precision, and font canonicalisation for block-nested MTEXT

KarimJerbi:
Fixes three compounding bugs in block-cached MTEXT rendering.

**Font resolution**
 `resolve_font` returned raw TTF stems (`arial`) to `Face::resolve`, which requires canonical family names (`Arial`).
Block-nested MTEXT with inline `\f`/`\F` overrides produced no geometry : Fixed by returning `Cow<'a, str>` and routing through
`sysfont::canonical_family_name()`, falling back to the style font on failure. Zero allocations on the common paths.

**Per-wire colour**
 `tessellate_sub_local` folded all tessellator output into one `LocalWire`, keeping only the first wire's colour.
Inline `\C`/`\c` overrides inside block definitions were silently discarded. Fixed by returning `Vec<LocalWire>`  one entry per tessellator wire.

**Fill geometry** two sub-issues:
- `fill_tris_low` (double-single precision low half) was never forwarded from `WireModel` to `LocalWire`, silently producing ~0.5 m quantization error on fill triangles at UTM-scale coordinates.  `debug_assert_eq!` added at the emit site to enforce the pairing invariant in debug builds.
- Fill-only wires (empty `points`, non-empty `fill_tris`) were routed through `emit_text_baseline`/`emit_greeked_text`, which ignore `fill_tris`. Fill never rendered. Fixed by bypassing the text LOD ladder for fill-only wires via `is_fill_only: bool` on `LocalWire` (also replaces three independent re-derivations of the same predicate, and fixes `StyleKey::fill_only` being hardcoded `false` at bot `emit_wire` call sites).

**Performance** 
profiled on a 93,509-entity production DWG:
1.1 avg wires/insert, no measurable regression. Synthetic worst-case (200× 5-colour `\C` block): 10.0 avg wires/insert, 380 ms expand (debug) correct, each colour segment produces outline + fill.

**Tests**
104 unit + 1 integration, all passing.

**Known debt**
6 sites in `dimension.rs`, `multileader.rs`, `table.rs` emit `fill_tris` without pairing `fill_tris_low`. Will trip the new `debug_assert_eq!` when those entities appear inside block definitions. FIXME comments added at each site.

various AIs used for debugging and review


KarimJerbi:
Should fix #177, more testing needed

PR #212, opened 2026-06-28
chore: disable printpdf default features (unused html deps)

KarimJerbi:
printpdf's default = ["html"] pulled in azul-layout, rust-fontconfig, kuchiki, xmlparser and their transitive trees (~40+ crates including two with cdylib targets) — none of which are used by the project's PDF export (only basic primitives like Mm, Line, Polygon, Rgb).

Also fixes a MinGW linker error (export ordinal too large: 100035) triggered when ld.bfd tries to build a DLL for printpdf with >65535 exported symbols.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,7 +80,7 @@ dependencies = [
  "encoding_rs",
  "flate2",
  "indexmap",
- "itoa 1.0.18",
+ "itoa",
  "nalgebra",
  "nom 7.1.3",
  "once_cell",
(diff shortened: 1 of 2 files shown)


HakanSeven12:
Merged into `main` via cherry-pick as 3d23c45 (authorship preserved). Verified: `pdf_export` builds clean with `default-features = false` (all primitives used — Op/PdfDocument/Line/Polygon/Rect/CurTransMat/Rgb — are core, none html-gated), and azul-layout/kuchiki/rust-fontconfig/xmlparser trace only to printpdf so nothing else breaks. Cargo.lock reconciled against main's acadrust patch and validated by a full build. Thanks!

PR #257, opened 2026-07-02
perf(core): cache ribbon_groups() with OnceLock, avoid per-frame tree allocations

KarimJerbi:
Change CadModule::ribbon_groups() from -> Vec<RibbonGroup> to -> &[RibbonGroup], backed by a per-module OnceLock cache. The ribbon group tree is static data (&'static str, Copy enums) that was being fully reconstructed — Vecs, String clones, enum discriminants — on every call, including all three per-frame call sites (view(), dropdown_overlay(), style_combo_overlay()).

- 7 built-in modules + 2 plugin templates cache via function-local OnceLock. Safe because each is a unit struct with exactly one instance per process (plugins run in isolated child processes; see PluginProcess::spawn()).
- SharedCadModule avoids the static pattern entirely, storing groups in an instance field (owned.rs).
- IPC runner converts &RibbonGroup -> OwnedRibbonGroup directly via new From<&T> impls, avoiding an intermediate clone.
- render_small/render_large and the two make_tool_row closures now borrow (&RibbonItem, &[ToolDef]) instead of taking ownership, so the view loop doesn't need to clone items to satisfy the old by-value signatures.
- Various match-ergonomics deref fixes (*id, *default, *icon, etc.) from the &RibbonItem pattern change.

### DIFF
--- a/crates/ocs_plugin_api/src/ribbon.rs
+++ b/crates/ocs_plugin_api/src/ribbon.rs
@@ -138,12 +138,113 @@ pub trait CadModule: Send + Sync {
     #[allow(dead_code)]
     fn id(&self) -> &'static str;
     fn title(&self) -> &'static str;
-    fn ribbon_groups(&self) -> Vec<RibbonGroup>;
+    fn ribbon_groups(&self) -> &[RibbonGroup];
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn once_lock_eliminates_allocation_after_first_call() {
+        // Helper to build a realistic module tree (2 groups, ~20 items)
+        // Benchmark builds this manually for a clean before/after comparison.
+        fn build_tree() -> Vec<RibbonGroup> {
+            vec![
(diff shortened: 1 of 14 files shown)


KarimJerbi:
## Per-frame budget at 60 fps

The ribbon render path calls `ribbon_groups()` three times per frame (`view()`, `dropdown_overlay()`, `style_combo_overlay()` all three hit the same OnceLock, so each is a pointer load after init:

| Metric      | Before (3× 3.1 µs) | After (3× 0.4 ns) |
|-------------|-------------------|-------------------|
| Per frame   | 9.3 µs            | 1.2 ns            |
| Per second  | 558 µs            | 72 ns             |
| Per minute  | 33 ms             | 4.3 µs            |

very small gains but this change closes off a legitimate tail-risk, if we later add more groups/tools, or the render path gets busier (more overlays open simultaneously, more redraws), the old code's cost scales with tree size x call frequency.

HakanSeven12:
thank you

Issue #258, opened 2026-07-03
Cursor movement is very slow in some file


jaisekjames:
Cursor movement is slow in some file.

[AEC Bldg Plan Sample.dwg.txt](https://github.com/user-attachments/files/29625457/AEC.Bldg.Plan.Sample.dwg.txt)

ceremcem:
> the performance issues come from render side. So file format not important. Can you create another issue and send your file? 

 _Originally posted by @HakanSeven12 in [#203](https://github.com/HakanSeven12/OpenCADStudio/issues/203#issuecomment-4876561830)_

I think resolving this issue will also resolve my case. I downloaded the provided file by the OP and it even behaves worse than with my file. The file I'm working on is proprietary so I'm not able to provide that specific file. But I *think* there is a slight improvement since 79a5ac6. I've compiled f4fffa1 and it feels a bit smoother.

HakanSeven12:
fixed. Thanks to @sLuCHaa

ceremcem:
I tested my build from commit b88e5e4 and zooming in and out out still flickers with my file. There is definitely an improvement, though. For demonstration, I copied the OP's drawing multiple times. Here is the resulting behavior: 

https://github.com/user-attachments/assets/e66e688e-e5d6-4ffc-94ed-ad1c29ca9351

The RAM usage at the last state was 2.2GB, FYI

sLuCHaa:
The flicker @ceremcem is seeing is the interaction-LOD from #273 doing its job a bit too visibly: to keep navigation fast on hatch-heavy drawings it suppresses the hatch fills while the view is moving and brings them back once it settles. On a slow wheel-zoom the individual ticks land around the settle window, so the fills toggle on/off — that's the flicker.

#300 fixes it at the source rather than tuning the LOD: each hatch boundary is triangulated once (even-odd, islands/holes included) and those triangles are rasterized directly, instead of drawing a screen-filling quad and running a point-in-polygon test per fragment. That makes the hatch pass cheap enough that the navigation LOD isn't needed at all, so the fills just stay visible while you zoom — no flicker.

The old LOD is still there behind OCS_HATCH_LOD (off by default) as a safety net. @ceremcem if you can build #300 and try it on your file, I'd love to know whether the flicker is gone on your end too.

ceremcem:
[...]

PR #273, opened 2026-07-05
perf(render): faster navigation on large drawings — scene-render cache + hatch interaction LOD

sLuCHaa:
Two related fixes for slow navigation on large drawings, likely relevant to #258 (cursor movement is very slow in some file). Both are on the render side, which matches @HakanSeven12's note there.

## 1. Scene-render cache — skip re-raster on unchanged frames

Every cursor movement re-rendered the entire GPU scene (all wires, hatches, meshes) even when nothing visible changed: iced redraws the whole window on each input message, so the shader's `draw` → `prepare` → `render` runs on every mouse move. On a large drawing one such redraw exceeds the frame budget, so moving the cursor stutters — and it scales with drawing size.

The pipeline already resolves each frame into a persistent resolve texture and blits that to the surface. This hashes the inputs that determine the rendered image (camera uniforms, geometry/selection generations, wire content id, render-mode flags, clip size, live preview) into a per-viewport signature. When a frame's signature matches the previous one the image is pixel-identical, so `render` skips passes 1–7 plus the MSAA resolve and re-blits the resolve texture instead. A pure cursor move becomes a single fullscreen blit instead of an O(N) re-rasterization.

Any real change (edit → `geometry_epoch`, select/hover → `selection_generation`, camera / render-mode → the hashed uniforms/flags) differs the signature and forces a full render, so nothing goes stale. The ViewCube is excluded (its own always-on pass) so cube hover keeps tracking while the scene is cached.

## 2. Hatch interaction LOD — skip the hatch pass while navigating

With the cache in place, a still view is fast but active pan/orbit/zoom was still slow on a hatch-heavy plan. Profiling (per-pass isolation) pinned it entirely on the batched hatch pass: it evaluates the fill procedurally per fragment — a point-in-polygon test over the whole boundary plus the pattern test, with a `discard` that disables early-Z. When the view sits inside a hatch, its quad covers the screen, so panning re-runs that full-screen shader every frame (~40 ms/frame here) while the wire pass and CPU prepare are negligible.

So while the view is actively panning / zooming / orbiting (detected from `camera_generation` + a short settle timer), the hatch pass is suppressed. That state is folded into the render signature, so the moment the view settles it renders hatches once and the cache (fix 1) holds that full-quality frame. A brief subscription keeps requesting frames just past the settle point so the hatched frame renders even when no input event would otherwise fire. The paper sheet keeps its fills.

## Result

On a hatch-heavy architectural plan (~260k wire points, tested on a discrete GPU): cursor movement over a static view went from stuttering to smooth (fix 1), and panning/zooming went from ~24 fps to smooth (fix 2), with hatches snapping back to full detail the instant the view settles.

## Testing

Built with `--release` and exercised on large and small DWGs: navigation is smooth, and edits, selection, hover highlight, layer/style changes and render-mode toggles all still update correctly (no stale frames).

### DIFF
--- a/src/app/view/mod.rs
+++ b/src/app/view/mod.rs
@@ -1462,6 +1462,16 @@ impl OpenCADStudio {
         } else {
             Subscription::none()
         };
+        // Interaction LOD: while the view is (or just was) navigating, keep
+        // requesting frames a little past the settle point. Panning itself is
+        // driven by input events, but once the cursor stops no event would fire
+        // the one full-quality frame that re-renders hatches — this tick does,
+        // then the scene-render cache holds it and the subscription auto-stops.
+        let nav_settle = if self.tabs[self.active_tab].scene.is_settling() {
+            window::frames().map(Message::Tick)
+        } else {
+            Subscription::none()
+        };
         // Blink the MText preview caret while the editor is open.
         let caret_blink = if self.mtext_editor.is_some() {
             iced::time::every(std::time::Duration::from_millis(530))
@@ -1493,6 +1503,7 @@ impl OpenCADStudio {
             history_tick,
             grip_dwell,
             hover_dwell,
+            nav_settle,
             caret_blink,
             web_fonts,
             autosave,
--- a/src/scene/mod.rs
+++ b/src/scene/mod.rs
@@ -901,6 +901,13 @@ pub struct Scene {
     /// (re)built, otherwise the held value. `build_primitive` reads it right
     /// after the call to gate GPU wire re-upload. 0 = none yet.
     pub(crate) last_model_wire_gen: std::cell::Cell<u64>,
+    /// Interaction-LOD state: `camera_generation` seen on the previous frame and
+    /// the wall time it last changed. Used to detect "the view is actively being
+    /// panned / zoomed / orbited" so the expensive per-pixel hatch pass can be
+    /// suppressed while moving and rendered once on settle (held by the
+    /// scene-render cache). See [`Scene::navigating_lod`].
+    nav_last_gen: std::cell::Cell<u64>,
+    nav_changed_at: std::cell::Cell<Option<iced::time::Instant>>,
     /// Static-hold cache for the Model layout: the FULL, un-culled tessellation
     /// held resident and reused for every camera (the geometry is
     /// camera-independent — see `model_tile_wires_arc`). `(epoch, gen, wires)`;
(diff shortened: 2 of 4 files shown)


semhustej:
I also encountered degraded performance in files, which are not that big. Every performance improvement is helpful.

Could this possibly solve https://github.com/HakanSeven12/OpenCADStudio/issues/258 ?

sLuCHaa:
Thanks! Yes, I think this directly targets #258 — that issue is exactly cursor-movement slowness, and it lines up with @HakanSeven12's note there that the performance comes from the render side.

Root cause this fixes: moving the cursor re-rendered the entire GPU scene every frame, even when nothing visible changed (iced redraws the whole window on each input event). On a large drawing one such redraw blows the frame budget, so the cursor stutters — and it scales with drawing size. With this change, a frame whose view / geometry / selection / preview are all unchanged skips every geometry pass and just re-blits the cached resolve texture, so a pure cursor move costs a single blit instead of re-rasterizing the whole drawing.

One honest caveat: this removes the wasted re-render on cursor-only movement. Active pan / orbit / zoom still renders every frame (the view genuinely changes), so if part of the slowness is during panning on very large files, that would need a separate follow-up (interaction LOD). But the "just moving the cursor is slow" symptom should be gone.

The AEC Bldg Plan Sample.dwg attached to #258 would be a good file to confirm against — happy if someone can try this branch on it.

sLuCHaa:
Quick follow-up on that caveat about panning — I went ahead and did the interaction-LOD pass and pushed it as a second commit here.

I profiled the panning case on the AEC Bldg Plan Sample.dwg and it turned out to be almost entirely the batched hatch pass. The hatch fill is evaluated procedurally per fragment (a point-in-polygon test over the whole boundary + the pattern test, with a discard that disables early-Z). When the view sits inside a hatch, its quad covers the whole screen, so panning re-runs that full-screen shader every frame — ~40 ms/frame on that file, while the wire pass and the CPU-side prepare were negligible by comparison. That's also why zooming in didn't help: you're still inside the hatch.

So now, while the view is actively panning / orbiting / zooming, the hatch pass is skipped, and the moment it settles it renders hatches once and the resolve-texture cache from the first commit holds that full-quality frame. On that file panning went from ~24 fps to smooth, with hatches snapping back to full detail as soon as you stop. 

So both the cursor-only stutter and the pan/zoom slowness should be covered now. Would still appreciate a second pair of eyes / a test on other files.

HakanSeven12:
It works great! Thank you

PR #292, opened 2026-07-06
fix(plot): correct hatch rendering in PDF / plot export

sLuCHaa:
## Summary

Hatched content plotted incorrectly in the PDF / plot export — especially fills nested inside a block INSERT (a title-block logo, for example): they were missing, mis-spaced, or drew phantom bars. This PR makes the PDF/plot output match.

## Fixes

**1. Export block-internal hatch fills**
The export path (`paper_canvas_hatches`) collected only hatches owned directly by the layout block, so a hatch nested in a block INSERT was dropped and the block printed as bare monochrome outlines. The insert-explosion the viewport already performs is extracted into a shared `exploded_insert_hatch_models()`, now called from both the viewport (`synced_hatch_models`) and the export — so a plot draws block-internal hatches identically to the screen.

**2. Honour the hatch's own stored pattern-line spacing**
`hatch_model_from_dxf` re-derived pattern spacing from the name-matched catalog entry × `pattern_scale`, ignoring the resolved line geometry the DWG stores on the hatch itself. When a drawing was authored against a different base spacing (e.g. imperial `0.125` vs the catalog's metric `3.175` for ANSI31), lines came out up to ~25× (inch→mm) too coarse and a dense fill collapsed to a few stray lines. When the hatch carries its own line geometry it is now used directly (identity scale/angle); the catalog lookup remains the fallback.

**3. Fill far-from-origin pattern hatches**
`pattern_segments` clamped the **absolute** scan-line index to `±MAX_LINES_PER_FAMILY`. A fine-spaced hatch far from the pattern origin has large-magnitude indices at both ends but a small span, so the clamp inverted the range (`k_lo > k_hi`) and emitted nothing — silently dropping the fill. It now caps the line **count** (span) instead of the absolute index.

**4. Skip TEXTBOX boundary paths**
TEXTBOX boundary paths (flag bit 3) are text bounding-boxes derives for island detection; they are never drawn or filled. Treating one as a fill boundary painted its rectangle solid — a phantom bar. Such paths are now skipped when building the fill boundary.

**5. Print white/ACI-7 hatch fills black on paper**
Hatch fills arrive adapted to the dark screen background, so a white/ACI-7 fill would vanish white-on-white on the sheet. The export now mirrors the wire pass (near-white/near-yellow → black, near-cyan → dark blue), matching colour-7-on-white plotting. Wipeouts keep their paper-white mask.

## Tests

Adds `tests/block_hatch_export.rs`:
- block-internal hatch reaches the export set,
- stored-line spacing is honoured (not re-scaled from the catalog),
- far-from-origin fills are not dropped,
- TEXTBOX paths are not filled.

All self-contained (no external drawing required).

## Scope

Export/plot path only. An unrelated on-screen GPU hatch-clipping nuance (dense pattern fills rendering over their bounding box in the viewport) is **not** part of this PR — it does not affect the plotted output.

### DIFF
--- a/src/io/pdf_export.rs
+++ b/src/io/pdf_export.rs
@@ -381,9 +381,28 @@ fn emit_hatch(ops: &mut Vec<Op>, hatch: &HatchModel, ox: f32, oy: f32) {
     if hatch.boundary.is_empty() {
         return;
     }
-    let [r, g, b, a] = hatch.color;
+    let [mut r, mut g, mut b, a] = hatch.color;
     if a < 0.01 {
         return;
+    }
+    // Adapt hatch fills to the white sheet, mirroring the wire pass: colours
+    // arrive adapted to the (dark) screen background, so a white/ACI-7 fill
+    // would vanish white-on-white on paper. Force near-white/near-yellow → black
+    // and near-cyan → dark blue, matching AutoCAD's colour-7-on-white plotting.
+    // Genuine colours are untouched; WIPEOUTS keep their paper-white mask.
+    if hatch.name != "WIPEOUT_FILL" {
+        let is_light = r > 0.80 && g > 0.80 && b > 0.80;
+        let is_yellow = r > 0.80 && g > 0.70 && b < 0.30;
+        let is_cyan = r < 0.30 && g > 0.70 && b > 0.70;
+        if is_light || is_yellow {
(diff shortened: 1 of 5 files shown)


HakanSeven12:
thank you

PR #293, opened 2026-07-06
feat(plot): snap plot-window corners to geometry endpoints

sLuCHaa:
## Summary

`PLOTWINDOW` / `PW` now snaps to geometry while you pick the two corners of the plot window, so the plotted area lands exactly on a frame's corners.

Previously the plot-window corners followed the raw cursor (the global object-snap master toggle is off by default, so nothing snapped), and it was hard to hit the frame corners precisely.

## Behavior

While `PLOTWINDOW` is picking either corner (model or paper space), **Endpoint** object-snap is forced on — independent of the global OSNAP toggle — with the normal snap marker, so the pick lands on frame corners. The snapper's master toggle and configured modes are saved before the pick and restored after, so the user's global OSNAP settings are unchanged. No other command is affected.

## Implementation

- `CadCommand::forces_endpoint_snap(&self) -> bool` — new opt-in trait method (default `false`); only `PlotWindowCommand` returns `true`.
- `Snapper::snap_forced_corners(...)` — same signature as `snap()`; replaces the active mode set with **Endpoint only**, runs the existing `snap()`, then restores `snap_enabled` and the mode set.
  - Endpoint only, deliberately: frame corners are line endpoints, so a single O(n) Endpoint pass catches them. Intersection snap is O(n²) over every segment pair in the hit-test wire set and froze the UI on large drawings viewed zoomed-out (exactly when a plot window is picked), so it is excluded.
- The cursor-move handler routes the plot-window pick through `snap_forced_corners` when the active command opts in, leaving every other command on the existing `snap()` path.

The rest of the snap pipeline (marker rendering, effective-point selection, `on_point`) is unchanged.

## Tests

- `PlotWindowCommand::forces_endpoint_snap()` returns `true` (opt-in).
- `forced_corner_modes()` is exactly `{Endpoint}` (no Intersection/Midpoint/Center).
- `snap_forced_corners` restores `snap_enabled` and the mode set after the call.

The live corner-snap-during-pick was verified manually (it depends on the camera projection, already covered by existing snap tests).

## Scope

`PLOTWINDOW` / `PW` only. No changes to global OSNAP defaults or any other snap caller.

### DIFF
--- a/src/app/update/viewport.rs
+++ b/src/app/update/viewport.rs
@@ -889,8 +889,19 @@ pub(super) fn on_tick(&mut self, t: Instant) -> Task<Message> {
                     } else {
                         let (go, gr) = self.tabs[i].ucs_grid_basis();
                         self.snapper.from_point = self.last_point;
-                        self.snapper
-                            .snap(snap_cursor, p, &all_wires[..], view_rot, eye, bounds, go, gr)
+                        let force_corners = self.tabs[i]
+                            .active_cmd
+                            .as_ref()
+                            .map(|c| c.forces_endpoint_snap())
+                            .unwrap_or(false);
+                        if force_corners {
+                            self.snapper.snap_forced_corners(
+                                snap_cursor, p, &all_wires[..], view_rot, eye, bounds, go, gr,
+                            )
+                        } else {
+                            self.snapper
+                                .snap(snap_cursor, p, &all_wires[..], view_rot, eye, bounds, go, gr)
+                        }
                     };
 
                     // Object Snap Tracking: update dwell, then align the cursor
--- a/src/command.rs
+++ b/src/command.rs
@@ -530,6 +530,13 @@ pub trait CadCommand: Send {
         false
     }
 
+    /// True if this command should have Endpoint/Intersection object-snap
+    /// forced on while it is active, regardless of the global OSNAP toggle.
+    /// Used by PLOTWINDOW so the plot-area corners always snap to geometry.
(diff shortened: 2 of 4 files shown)


HakanSeven12:
Hi. I reverted this pr. Why we want to force snap when user disabled the osnap mode?

sLuCHaa:
Fair — forcing snap while the global OSNAP is off overrides a deliberate user setting and special-cases one command, so the revert makes sense. There's also a concrete downside I missed: because the forced pick ignores the toggle, there's no way to place a corner off geometry (a margin, or a window that isn't aligned to the frame) — so it's strictly worse for that case.

The motivation was just that OSNAP is off by default, so picking a plot window against a frame is fiddly. But that's a weak reason to override one command.

If the underlying need is worth addressing, two convention-aligned options:

Leave PLOTWINDOW on the normal snap path (respect running OSNAP) — no special-casing.
An explicit opt-in preference ("snap plot window to geometry", default off), so it's the user's choice, not a silent override.
I'll leave it reverted. If you'd take #2, I can open a smaller PR for it — otherwise happy to drop it.

HakanSeven12:
osnap can be toggled by clicking F3 and most of the time user already activated the endpoint snap. So I assume this change is unnecessary :) thank you anyway

JuanJoseJimenezdeCisneros:
Hi, I have seen that when drawging entities, you have de option to Shift+right buttom and there are different snap options. I want an option "Medium point between two points". It is very useful.

Thank you.

JJJ

PR #300, opened 2026-07-07
perf(hatch): tessellate fills so navigation no longer flickers (#258)

sLuCHaa:
Fixes the remaining hatch flicker reported on #258 — fills toggling on/off while zooming on hatch-heavy drawings.

## Root cause
PR #273 made navigation fast on hatch-heavy drawings by suppressing the hatch pass while the view is navigating (interaction LOD), holding the full-quality frame in the scene-render cache once it settles. That suppression is what flickers: discrete mouse-wheel zoom ticks land around the 130&nbsp;ms settle window, so hatches toggle off (moving) / on (settled) / off… as you zoom.

The reason the LOD exists is the hatch pass itself. The batched shader draws a screen-covering AABB quad per hatch and, for every fragment, runs a point-in-polygon ray-cast over the whole boundary plus the pattern test, with a `discard` that disables early-Z. Zoomed inside a hatch the quad covers the screen, so the whole shader re-runs per pixel every frame (~40&nbsp;ms/frame in #273's profiling).

## Fix
Attack the source instead of the symptom: triangulate each hatch boundary once (lyon `FillTessellator`, even-odd rule — same coverage as the old `in_polygon`, islands/holes included) and rasterize those triangles. That removes both dominant costs — the screen-filling overdraw and the per-fragment boundary test — so the hatch pass is cheap enough that no interaction LOD, and therefore no flicker, is needed.

- Tessellation is cached with the batch build (only re-runs on `geometry_epoch`), in local (world_origin-relative) space, so the double-single relative-to-eye precision is unchanged.
- The fragment shader keeps `in_polygon` only on a per-instance fallback path (`poly_test == 1`) for boundaries lyon can't tessellate — those emit the old AABB quad and render identically.
- The interaction LOD is retained behind `OCS_HATCH_LOD` (default **off**) as a safety net; it can be removed in a follow-up once confirmed unnecessary on real files.
- The scene-render cache from #273 is untouched.

## Testing
- Unit tests for the tessellator (`hatch_tess`): square, even-odd hole, nested island (hole-in-hole → filled centre), self-intersecting boundary, degenerate → empty.
- Manual on the `AEC Bldg Plan Sample.dwg` from #258: slow and fast wheel zoom — fills stay continuously visible (no flicker); pattern / solid / gradient / island hatches render identically to before (compared against `OCS_HATCH_LOD=1`, the old path).

### DIFF
--- a/src/scene/mod.rs
+++ b/src/scene/mod.rs
@@ -1198,6 +1198,16 @@ impl Scene {
             .map_or(false, |t| t.elapsed().as_millis() < Self::NAV_SETTLE_MS)
     }
 
+    /// Whether the interaction-LOD hatch suppression is enabled (env
+    /// `OCS_HATCH_LOD`), read once. Default OFF: the tessellated hatch pass is
+    /// cheap enough that suppression — and its zoom flicker (#258) — is not
+    /// needed. Kept behind the flag as a safety net for pathological drawings.
+    pub fn hatch_lod_enabled(&self) -> bool {
+        use std::sync::OnceLock;
+        static ON: OnceLock<bool> = OnceLock::new();
+        *ON.get_or_init(|| std::env::var_os("OCS_HATCH_LOD").is_some())
+    }
+
     /// True for a short window around navigation — used by the subscription to
     /// keep requesting frames just past the settle point so the one full-quality
     /// (hatched) frame actually renders after the cursor stops, even when no
--- a/src/scene/pipeline/hatch_batched_gpu.rs
+++ b/src/scene/pipeline/hatch_batched_gpu.rs
@@ -5,7 +5,7 @@
 //
 // Data layout — three storage buffers fed from `HatchModel`s:
 //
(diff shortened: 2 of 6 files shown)


HakanSeven12:
I wanted to do somthing similar when I finished text rewrite :) thank you

HakanSeven12:
Merged into `main` via fe00bc3 — thanks @sLuCHaa! 🎉

Reviewed: triangulating each boundary once (cached with the batch build) + rasterizing the triangles removes both the screen-fill overdraw and the per-fragment `in_polygon` raycast, so the flicker-causing interaction LOD is no longer needed (kept behind `OCS_HATCH_LOD`). Precision is preserved — the triangles stay in the same world_origin-relative frame and go through the identical double-single RTE path as the old AABB corners — and untessellatable boundaries fall back to the old quad + `in_polygon` path via `poly_test`. No new dependency (`lyon_tessellation` was already in the tree). Builds clean, all tests green alongside the SDF-text work already on main.

Issue #161, opened 2026-06-22
COPY do not copy Dimensions


pacaeiro:
Copy command do not copy dimensions.

P.S. - I know I'm resuming very much in the reports.
If you want me to put more info, just say.

Thank you!

HakanSeven12:
fixed and will be available on 0.6.7. if problem still exist please record a video

PR #303, opened 2026-07-07
fix(clipboard): paste copies dimensions correctly (#290)

### LINKED ISSUE #290: Dimension copy-paste error (CTRL+C / CTRL+V)
I would like to report another issue regarding the standard Copy-Paste (CTRL+C and CTRL+V) commands. Currently, when I select and copy objects, the Dimension (Dim) objects are not being copied along with them as they normally should. While other entities paste fine, the Dimensions are missing from the destination. 

https://github.com/user-attachments/assets/0b4333e5-4d21-4c47-aab2-4784cc9d1bae

sLuCHaa:
Fixes #290 — copy-paste (CTRL+C / CTRL+V) drops dimensions: other entities paste fine but dimensions are missing at the destination.

## Root cause
A dimension draws from its baked `*D` block (extension/dim lines, arrows, text MText), whose geometry is baked in **absolute WCS coordinates**. The in-drawing COPY gives a copied dimension its own transformed copy of that block and re-points `block_name` (the #161 fix). The clipboard **paste** path (`finalize_paste`) never did this — it added the dimension entity but left `block_name` pointing at the source block. So the pasted dimension:

- **same drawing:** renders from the original block at the *original* location, not the paste point → looks missing;
- **cross drawing:** the source block doesn't exist in the target → no baked geometry.

## Fix
- **Copy:** `ClipboardDeps::capture` now snapshots each copied dimension's `*D` block into the clipboard (`dim_blocks`), so the geometry travels with the paste — cross-drawing included.
- **Paste:** `finalize_paste` builds a fresh copy of that block, transformed by the paste offset, and re-points the pasted dimension's `block_name` at it — for same- and cross-drawing pastes alike. No orphan blocks (the snapshot is only cloned when a dimension is actually pasted).
- Refactor, no behaviour change: the block-building is factored out of `clone_transformed_block` into a shared `define_transformed_block`, and the per-block snapshotting out of `capture_blocks` into `snapshot_block`.

## Testing
Manual (native): a drawing with dimensions — select + copy + paste at a new point; the dimension now appears at the paste location, rendered identically. In-drawing COPY of a dimension still works. Also verified pasting into another tab.

### DIFF
--- a/src/app/command_driver.rs
+++ b/src/app/command_driver.rs
@@ -2012,6 +2012,35 @@ impl OpenCADStudio {
                 if let Some(t) = &translate {
                     crate::scene::view::dispatch::apply_transform(&mut entity, t);
                 }
+                // A dimension draws from its baked `*D` block (baked in WCS), so
+                // give the paste its own transformed copy of that block and
+                // re-point it — otherwise the pasted dimension renders at the
+                // source location instead of the paste point. The block was
+                // snapshotted into the clipboard at copy time, so this works
+                // cross-drawing too. Mirrors the in-drawing copy. (#290, #161)
+                if let acadrust::EntityType::Dimension(d) = &entity {
+                    let bn = d.base().block_name.clone();
+                    if !bn.trim().is_empty() {
+                        let subs = self
+                            .clipboard_deps
+                            .dim_blocks
+                            .iter()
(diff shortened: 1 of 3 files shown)


HakanSeven12:
thank you

Issue #302, opened 2026-07-07
Could we list OCS version in the "manage" tab"


KevinGriffin-new:
For diagnostics and user experience it might be good to have an "about" tab under "manage" that lists ocs version number. I know it's available through the command "about" , but it might be nice to have a fallback

HakanSeven12:
Resolved by #306 (merged to main, commit efa4bdd). ABOUT now has a ribbon entry in the Manage tab → Application group.

KevinGriffin-new:
Thank you!

PR #306, opened 2026-07-07
feat(manage): surface About in the ribbon — Application group (#302)

KevinGriffin-new:
The ABOUT command and its version-reporting window already exist, but the ToolDef sat dead-code in the draw module with no ribbon entry, so the only way in was typing ABOUT. Move it to the manage module and give it an Application group on the Manage tab, as a discoverable fallback for version diagnostics.


HakanSeven12:
thank you

PR #317, opened 2026-07-08
Bug fixes (text/dimension), SVG export, solid persistence, CI & tests

mf4633:
## Description
A set of fixes, tests, and one new feature developed on the mf4633 fork and verified green (test suite grown 38 → 101). Two are user-visible correctness bugs in text and dimension rendering; one adds SVG export; the rest cover persistence, performance, CI, and test coverage. Everything was verified headlessly (build + tests + clippy, and an actual SVG render); GUI paths were exercised only where drivable without a display, so a quick interactive smoke-test is worthwhile.
## Type of change
- [x] Bug fix (non-breaking)
- [x] New feature (non-breaking)
- [x] Tests / CI / tooling
- [ ] Breaking change
## Changes
Each group is independent and could instead be taken as its own PR:
### 1. Bug fixes
- **Text width factor applied twice.** `lff::tessellate_text_run` multiplied the glyph advance by the width factor and then multiplied the cursor by it again (wf^2), so a 2x factor spread glyphs ~3.5x and over-ran the width MTEXT layout allotted. Now the cursor advances in glyph units and wf is applied once, matching `text_local_bounds` / `measure_word`.
- **Dimension formatters dropped rounding carries:**
  | Input | Was | Now |
  |-------|-----|-----|
  | `format_dms(0.99999deg)` | `0d59'60"` | `1d0'0"` |
  | `format_engineering(11.999")` @2dp | `0'-12.00"` | `1'-0.00"` |
  | `format_architectural(6.999")` | `0'-6 1"` | `0'-7"` |
  Each component is rounded then carried (sec->min->deg, in->ft, frac->in->ft).
### 2. SVG export (SVGOUT / EXPORTSVG)
Vector export of the 2D drafting view. Tessellated WireModel sub-strokes -> <polyline>s, coloured per wire, fit to a viewBox, Y mirrored (CAD Y-up -> SVG Y-down), WYSIWYG page background. Command wiring mirrors STLOUT.
### 3. Persistence -- EXTRUDE / REVOLVE / SWEEP / LOFT / IMPORTOBJ
These stored their solid under a placeholder empty Solid3D, so it vanished on save/reload. The tessellation is now persisted as an ACAD_MESH entity (round-trips through both DWG and DXF) and re-tessellated to a shaded mesh on load (new scene::mesh_tess). Also removes the dead CommitSolid3D path (primitives already persist via real ACIS).
### 4. Performance, CI & tests
- Lazy raster-image decode (roadmap 1.6): decode deferred to first GPU upload via Arc<OnceLock>; off-screen images never decode.
- Profiling spans (roadmap 5.1): puffin behind a --features profile flag (zero cost otherwise).
- CI: build + test + clippy on every push/PR (was release-only); fmt non-blocking to respect the manual formatting.
- Tests 38 -> 101 on previously-untested code: offset/fillet/trim helpers, the LFF text engine, and the dimension formatters.
- Roadmap notes corrected (1.6 / 5.1 done; 3.5 and primitive-persistence already satisfied).
## How has this been tested?
- \`cargo build --all-targets\`, \`cargo test\` (101 pass), \`cargo clippy\`, and \`cargo check --features profile\` all green.
- Persistence: an integration test saves an ACAD_MESH entity and reads it back through the real acadrust writers/readers for both DWG and DXF.
- SVG export: unit tests plus an end-to-end headless-Chromium render of a sample drawing (correct geometry, orientation, colours).
- Bug fixes: regression tests assert linear width-factor scaling (and parity with the layout path) and each dimension rounding carry.
## Checklist
- [x] Builds without warnings on stable Rust
- [x] cargo test passes (38 -> 101)
- [x] cargo clippy clean on the changed code
- [x] MSRV preserved (Rust 1.75+)
- [x] README updated (SVG export added to the format list)
- [ ] Interactive GUI smoke-test by a maintainer with a display

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,65 @@
+name: CI
+
+# Build, test, and lint on every push to main and on every pull request.
+# release.yml only runs when a release is published, so nothing guarded the
+# build between releases — a compile break or a failing test could sit on
+# main unnoticed. This workflow closes that gap.
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
+# Cancel an in-progress run when a newer commit is pushed to the same ref.
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build-and-test:
+    name: Build & test (Linux)
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Rust (with clippy + rustfmt)
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy, rustfmt
+
+      # Same system libraries the release AppImage build installs — iced /
+      # wgpu / fontconfig need them to link, even for `cargo test`.
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update -q
+          sudo apt-get install -y \
+            libgl1-mesa-dev libx11-dev libxcursor-dev libxi-dev libxrandr-dev \
+            libxkbcommon-dev libwayland-dev libfontconfig1-dev libfreetype6-dev
+
+      - name: Cache cargo registry and build artifacts
+        uses: Swatinem/rust-cache@v2
+
+      - name: Build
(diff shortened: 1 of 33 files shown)


HakanSeven12:
looks like your fork too old. Can you update and compere?

HakanSeven12:
Im closing this one for now. You can open a new pr when you finish the migration

PR #322, opened 2026-07-08
perf: keep large copy/paste and point picking responsive on big drawings

sLuCHaa:
## Summary

On large drawings, copying/pasting a big selection — and even just picking a
point (MOVE, plot window, …) with object snap on — could freeze the app until it
had to be force-quit. This PR removes the algorithmic hot spots that scaled with
the whole document instead of the selection or the cursor neighbourhood, so
large-drawing editing stays responsive.

Everything here is a self-contained, low-risk change; no file formats, no
rendering pipeline changes.

## What was slow, and what changed

Measured on a ~250k-entity drawing (M3, release build):

| Area | Before | After | Fix |
|---|---|---|---|
| Point picking with OSNAP (MOVE/paste placement) | freeze (∞) | ~0.1–4 ms/move | gate the O(k²) intersection passes |
| Copy anchor (COPYCLIP centroid) | ~256 ms | ~3 ms | bounding-box mean instead of tessellating everything |
| Properties panel rebuild (huge selection) | ~4.5 s | ~70 ms | count-only panel above a cap |
| Paste placement ghost | freeze | smooth | budget the per-frame ghost by wire count |
| Paste commit (mesh rebuild) | full-doc re-tessellation | incremental | only tessellate newly added solids |

### Commits

- **perf(snap): keep point picking responsive on large drawings**
  Object snap re-scanned the whole drawing every cursor move, and the
  Intersection / Apparent-Intersection passes compare *pairs* of in-range
  segments — O(k²) in the number of wires whose bounding box overlaps the snap
  aperture. Zoomed out (or with long lines whose AABB spans the view) that grows
  to the whole document and hangs. Now the in-range wires are counted once and
  only the pairwise passes are gated on it; every single-wire snap (endpoint,
  midpoint, centre, perpendicular, nearest) still runs, so corners still snap at
  any zoom.

- **perf(properties): count-only panel for very large selections**
  The panel aggregated shared/varying values across the selection: O(n) per row
  plus an O(n²) group filter (`group.handles.contains` is a linear scan per
  entity). Above a cap it now shows a lightweight `<n> objects selected` panel;
  bulk layer/colour/lineweight edits still go through the ribbon.

- **perf(clipboard): derive the paste anchor from bounding boxes**
  COPYCLIP/CUTCLIP tessellated every selected entity and averaged all wire
  vertices just to find the paste anchor. The anchor only decides which point of
  the selection sits under the cursor, so the mean of each entity's bounding-box
  centre is an equally valid, far cheaper origin.

- **perf(clipboard): cap the ghost preview by wire count, box beyond it**
  The paste ghost is cloned and re-uploaded on every cursor move, so its cost
  scales with the wire count. A whole-drawing paste flooded the event loop. The
  budget now also considers the wire count and falls back to a single
  bounding-box outline above it — the full-geometry ghost still shows for typical
  pastes (up to ~20k wires); only a whole-drawing paste rubber-bands a box.

- **perf(clipboard): keep large copy/paste responsive** (earlier commit)
  O(H·N) centroid scan replaced with the document's O(1) handle index;
  incremental `populate_missing_meshes_from_document` on paste instead of a
  full-document solid re-tessellation.

## Known limitation / possible follow-up

The paste ghost falls back to a **bounding box** for whole-drawing pastes
(> ~20k wires) rather than showing the full geometry, because the preview is
re-uploaded to the GPU every cursor move — O(wires) per frame.

The "right" fix
would be to upload the ghost geometry **once** and translate it per frame via a
small offset uniform. The current renderer bakes absolute world positions into
the wire buffers (double-single, reconstructed relative-to-eye in the shader) and
has no per-object model transform, so this needs a new offset uniform + bind
group added to the wire pass (and both the native and wasm shader variants). That
felt like a larger, separate change best reviewed on its own — happy to open a
follow-up PR if you'd like it.

Similarly, undo currently snapshots the whole `CadDocument` per edit; a delta
undo (record the added handles) would scale better in time and memory on very
large drawings. Also out of scope here.

## Testing

- `cargo build --release` — clean.
- `cargo test --lib` — 159 pass; the one failing test (`scene::text::lff …
  Turkish 'Ğ' fallback`) is unrelated to this change (font fallback) and fails on
  the base as well.
- Manually exercised copy / cut / paste and MOVE with OSNAP on a ~250k-entity
  drawing: no freeze, snapping still lands on corners/endpoints.

### DIFF
--- a/src/app/command_driver.rs
+++ b/src/app/command_driver.rs
@@ -2080,7 +2080,12 @@ impl OpenCADStudio {
             })
             .collect();
         self.merge_clipboard_ext_objects(i, &by_index);
-        self.tabs[i].scene.populate_meshes_from_document();
+        // Incremental: `add_entity` already tessellated every pasted top-level
+        // solid, and existing document solids are still cached — so only newly
+        // introduced block-definition solids need building. The full rebuild
+        // would clear and re-tessellate the entire document (every solid in the
+        // drawing) on each paste, which is what made a large paste stall.
+        self.tabs[i].scene.populate_missing_meshes_from_document();
         by_index
     }
 
--- a/src/app/commands/blocks.rs
+++ b/src/app/commands/blocks.rs
@@ -70,8 +70,9 @@ impl OpenCADStudio {
                         .iter()
                         .filter_map(|&h| self.tabs[i].scene.document.get_entity(h).cloned())
                         .collect();
-                    self.clipboard_centroid = super::super::helpers::entities_centroid(
-                        &self.tabs[i].scene.wire_models_for(&handles),
+                    self.clipboard_centroid = super::super::helpers::entities_centroid_by_bbox(
+                        &self.tabs[i].scene.document,
+                        &handles,
                     );
                     self.clipboard = entities;
                     self.clipboard_deps = super::super::ClipboardDeps::capture(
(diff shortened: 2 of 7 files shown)


ceremcem:
Wow

HakanSeven12:
thank you

PR #323, opened 2026-07-08
fix(hatch): render baked patterns with many line families (#313)

### LINKED ISSUE #313: Hatch dsplayed incorrectly
OCS  result
Image

ODA result

Image

sLuCHaa:
Fixes #313.

## Problem
Hatch fills are rasterised on the GPU by a per-fragment shader that holds at most `MAX_FAMILIES` (16) pattern line families. Other CAD software bakes complex patterns (`ROOFING`, `GRAVEL`, stone) into dozens–hundreds of one-off line families, so their fill rendered truncated — the roof shingle pattern in the sample showed only its horizontal courses, with every vertical joint missing (see the issue's OCS-vs-ODA comparison).

## Fix
Route hatches whose pattern exceeds the shader's family limit through the existing CPU rasteriser (`HatchModel::pattern_segments`, already used for PDF export and paper space) and draw the result as line-segment wires, which have no family cap. Simple patterns keep the fast GPU path; the GPU fill is skipped for the offloaded ones so nothing is drawn twice.

## Changes
- `HatchModel::needs_wire_offload()` — pattern has more families than the shader holds.
- `tessellate` — such hatches emit their pattern as wire segments instead of `vec![]`.
- `upload_hatches` — skips the GPU fill for offloaded hatches.

### DIFF
--- a/src/scene/convert/tessellate.rs
+++ b/src/scene/convert/tessellate.rs
@@ -126,7 +126,16 @@ pub fn tessellate(
     // dominant wire-instance cost on hatch-heavy drawings (issue #131). Picking
     // is unaffected: hatches are caught by their fill area through the existing
     // `pick::hit_test::click_hit_hatch` path, not this outline.
-    if matches!(entity, EntityType::Hatch(_)) {
+    //
+    // A pattern with more line families than the GPU shader holds renders
+    // truncated as a fill, so rasterise it to line wires on the CPU instead
+    // (the GPU fill is skipped for these in `upload_hatches`). (#313)
+    if let EntityType::Hatch(dxf) = entity {
+        if let Some(model) = crate::scene::Scene::hatch_model_from_dxf(dxf, entity_color) {
+            if model.needs_wire_offload() {
+                return hatch_offload_wires(&model, handle, color);
+            }
+        }
         return vec![];
     }
 
@@ -949,6 +958,27 @@ pub(crate) enum ArrowKind {
     Origin { size: f32 },
     Box_ { size: f32, filled: bool },
     Datum { size: f32, filled: bool },
+}
+
+/// Rasterise a many-family hatch pattern to line-segment wires (issue #313).
+/// `pattern_segments` already returns segments in the offset-relative WCS, so
+/// pack them into one WireModel with NaN sentinels between disjoint segments.
+fn hatch_offload_wires(
+    model: &crate::scene::model::hatch_model::HatchModel,
+    handle: Handle,
+    color: [f32; 4],
+) -> Vec<WireModel> {
+    let segs = model.pattern_segments();
+    if segs.is_empty() {
+        return vec![];
+    }
+    let mut pts: Vec<[f64; 3]> = Vec::with_capacity(segs.len() * 3);
+    for seg in &segs {
+        pts.push([seg[0][0] as f64, seg[0][1] as f64, 0.0]);
+        pts.push([seg[1][0] as f64, seg[1][1] as f64, 0.0]);
+        pts.push([f64::NAN, f64::NAN, f64::NAN]);
+    }
+    vec![WireModel::solid_f64(handle.value().to_string(), pts, color, false)]
 }
 
 pub(crate) fn arrow_from_block(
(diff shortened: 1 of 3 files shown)


HakanSeven12:
Hi 0.7.8 already solves this.

image

sLuCHaa:
Reopening-worthy, I think — but your call. I went back and measured this properly, and I owe you a correction first.

**My issue text was wrong.** I wrote that the vertical roof joints disappear. They don't. `ROOFING`'s first 16 pattern lines already contain both 0° and 90° families, so both directions still render. Your screenshot is correct. That is exactly why this looks fine at a glance.

**But the cap is real.** `upload_hatches` uploads at most 16 line families and drops the rest, with no CPU fallback:

- `src/scene/pipeline/hatch_gpu.rs:14` — `pub const MAX_FAMILIES: usize = 16;`
- `src/scene/pipeline/hatch_gpu.rs:422` — `families.iter().take(MAX_FAMILIES)`

Other CAD software bakes these patterns as one family per individual line: `ROOFING` = 65 families, `GRAVEL` = 41, `CAROLINA-LEDGESTONE` = 208.

I rasterised the sample drawing's hatches twice through the existing CPU rasteriser (`HatchModel::pattern_segments`, which `io/pdf_export.rs` already uses) — once with the full family list, once truncated to `MAX_FAMILIES`, i.e. exactly what the shader receives:

| Pattern | Families | Segments drawn |
|---|---|---|
| ROOFING | 65 | 500 / 2108 |
| GRAVEL | 41 | 445 / 1099 |
| CAROLINA-LEDGESTONE | 208 | 7 / 142 |

Nine hatches in the drawing exceed the limit. Across them the viewport draws **1195 of 4342 pattern segments**. The stone veneer keeps 15 of 255 — about 6% of its pattern. Note that PDF export already draws these in full, so the viewport and the export disagree today.

I've rebased the branch onto current `main` and added two tests that pin the argument down rather than asking you to trust my numbers:

- the cap loses geometry (16 of 64 lines for a synthetic baked pattern);
- the lines it keeps are a **strict subset** of the full rasterisation — nothing is moved, only omitted. That is what makes the wire offload purely additive: it cannot regress a hatch that already renders correctly.

Patterns within the limit (`AR-CONC` 13 families, `AR-SAND` 4, `ANSI31` 1) never touch the new path.

Happy to reopen if you'd like to take another look — or to close it for good if you'd rather not carry a second hatch path for nine hatches.

HakanSeven12:
Thanks for digging in — and for the honest correction on the issue text.

Quick clarification on the cap: `MAX_FAMILIES` is real, but it lives in the **legacy** per-primitive renderer, which today only draws **wipeout masks** (solid fills, zero pattern families) — so the cap never truncates anything. The active viewport hatch renderer is the batched path, which uploads every family (`family_count`, no cap) and loops over all of them in the fragment shader. That is why the complex baked patterns render in full and agree with the PDF export, and why the roof looked correct.

The confusion was on us: the uncapped renderer was named `hatch_batched_gpu` while the capped legacy one kept the `hatch_gpu` / `MAX_FAMILIES` name. I just renamed them on `main` (d37aacd) so the real renderer is `hatch_gpu.rs` (no cap) and the legacy wipeout-only one is `wipeout_gpu.rs` — `MAX_FAMILIES` now lives only there.

So no second hatch path is needed; keeping this closed. Appreciate the thorough measurements.

sLuCHaa:
You're right, and the rename makes the trap obvious in hindsight.

My mistake was mechanical: I grepped `.take(MAX_FAMILIES)` in `hatch_gpu.rs` and concluded the render path caps families, without ever checking that `upload_hatches` builds `HatchBatchedGpu`, and that `HatchGpu` now only serves wipeouts. The comment saying so was right above the line I edited. So my segment counts quantified a truncation that never happens at runtime.

For anyone finding this later: `hatch_batched.wgsl` reads `families` from an unbounded storage buffer and loops `i < inst.family_count`, so baked patterns render in full and match the PDF export.

What this branch actually did was pull nine hatches out of the correct renderer and draw them as CPU wires, which also bypasses the deliberate 2 px pattern LOD. That's a regression. Closing was the right call — thanks for the clear explanation, and sorry for the noise.

HakanSeven12:
No problem 🙂

PR #367, opened 2026-07-12
fix(ribbon): highlight toggle buttons by their actual state

KarimJerbi:
The UCS Icon, Properties, File Tabs, and Layout Tabs buttons were highlighted unconditionally because their ids fell through to an active_tool equality check in is_active_tool, which activate_tool sets on every click and never clears. Each id is now mapped to its corresponding visibility flag (show_ucs_icon, show_properties, show_file_tabs, show_layout_tabs), and those flags are kept in sync from their toggle handlers so the highlight reflects live state. The per-toggle booleans previously threaded through the render path are consolidated into a single ToggleState struct read from one place, so introducing a further toggle no longer requires changing every render-function signature and call site.

### DIFF
--- a/src/app/commands/display.rs
+++ b/src/app/commands/display.rs
@@ -81,6 +81,7 @@ impl OpenCADStudio {
                         let at_origin = sub == "ORIGIN";
                         // Update model-space icon flags.
                         self.show_ucs_icon = visible;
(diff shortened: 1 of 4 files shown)


HakanSeven12:
thank you

PR #383, opened 2026-07-14
fix(plot): never fail silently on PDF export; add EXPORTPDF <path> (#369)

### LINKED ISSUE #369: Export to PDF is broken
Export to PDF seems to be doing nothing (no errors and no action). No additional output in the console. Build commit 8b133ce

KevinGriffin-new:
## Problem (#369)

"Export to PDF seems to be doing nothing (no errors and no action). No additional output in the console."

Every code path in the export flow pushes a visible command-line message — except one. The flow hands off to an async `rfd` save dialog, and when that future resolves to `None` the app mapped it straight to `Task::none()`:

```rust
Message::PlotExportPath(None) => Task::none(),
```

`None` doesn't only mean the user cancelled. On Linux, rfd 0.17's XDG portal `SaveFile` call can fail outright (portal missing/broken under some desktops); rfd then falls back to zenity, and when that isn't installed either, the future still just resolves to `None` (see PolyMeilex/rfd#315 for what that failure mode looks like). rfd reports all of this only through the `log` crate — and OCS initializes `env_logger` only when `--log`/`RUST_LOG` is set — so the user sees exactly what the issue describes: no dialog, no error, no console output, no file. (The reporter runs a native Linux build from source, per their earlier issues.)

## Fix

- **`EXPORT` / `EXPORTPDF` now take an optional path** — `EXPORTPDF out.pdf` exports straight to the file with no dialog. This is both the workaround for systems where the native save dialog cannot open, and a headless-testable entry point. Relative paths land next to the drawing; a missing extension gets `.pdf`. (Same command-argument pattern as `CUIEXPORT <path>`.)
- **The dialog-`None` arms (`PlotExportPath` / `PlotWindowExportPath`) now push a command-line message** pointing at `EXPORTPDF <path>` instead of doing nothing, so the failure is never invisible again.
- The success message prints the full output path instead of just the file name.
- `COMMANDS.md` updated for the `[path]` argument.

@ceremcem — could you retry on this branch? If the save dialog still doesn't appear on your system, `EXPORTPDF <path>` should export regardless, and running with `--log warn` will show rfd's portal/zenity errors so we can see exactly why the dialog fails there.

## Verification

- New test `exportpdf_with_explicit_path_writes_pdf_without_dialog` drives `EXPORTPDF <path>` through the real dispatcher and asserts a valid `%PDF` file plus the confirmation message; `cancelled_pdf_save_dialog_reports_on_command_line` asserts the dialog-`None` feedback. **Both fail on the previous code** (silent no-op / unknown command).
- Full suite green: `cargo test --lib` — 196 passed.
- End-to-end on the real binary (`--serve`): `LINE` + `CIRCLE` then `EXPORTPDF <path>` writes a valid PDF with the expected geometry (visually checked).

Fixes #369

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -303,8 +303,8 @@ Status of every standard CAD command in Open CAD Studio:
 | `QSAVE` | — | Quick save | ✅ |
 | `SAVEALL` | — | Save all open drawings | ✅ |
 | `PLOT` | — | Print / plot | ✅ |
-| `EXPORT` | — | Export | ✅ |
-| `EXPORTPDF` | — | Export to PDF | ✅ |
+| `EXPORT [path]` | — | Export to PDF (save dialog, or straight to `path`) | ✅ |
+| `EXPORTPDF [path]` | — | Export to PDF (save dialog, or straight to `path`) | ✅ |
 | `PAGESETUP` | — | Page setup | ✅ |
 | `PLOTSTYLE` | — | Plot style | ✅ |
 | `PURGE` | PU | Purge unused items | ✅ |
--- a/src/app/automation.rs
+++ b/src/app/automation.rs
@@ -542,6 +542,56 @@ mod tests {
     }
 
     #[test]
+    fn exportpdf_with_explicit_path_writes_pdf_without_dialog() {
+        // EXPORTPDF <path> must export straight to the given file — no save
+        // dialog — and confirm on the command line. This is the dialog-free
+        // export path added for #369 (Export to PDF silently did nothing when
+        // the native save dialog could not open).
+        let mut app = OpenCADStudio::new_for_test();
+        app.automation_op(r#"{"op":"new"}"#);
+        app.automation_op(r#"{"op":"run","cmd":"LINE 0,0 100,50"}"#);
+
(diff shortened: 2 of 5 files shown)


JuanJoseJimenezdeCisneros:
Hi, I don't know if possible to import a pdf file and convert it into dwg or dxf entities.

KevinGriffin-new:
> Hi, I don't know if possible to import a pdf file and convert it into dwg or dxf entities.

That's an interesting problem. To do it correctly, layers, optical character recognition, rotated text read, takes a fair bit of work. If you're lucky, all that information is a available as vector data in the pdf. I'd recommend looking at third party tools or opening a new issue: https://github.com/HakanSeven12/OpenCADStudio/issues/new . On a side note I have been working on something like this but focused only on land survey pdfs, and pretty much only on subdivisions. If you @JuanJoseJimenezdeCisneros find a pdf to dwg converter that you like, please file a new issue, I'll have a look to see how it works. Thanks!

JuanJoseJimenezdeCisneros:
Autocad can import pdf files. I don't mean any type of pdf to convert into lines. I would be happy with next:
- Vector pdf files: convert into dxf/dwg entities
- Other components: insert as an image.

PR #384, opened 2026-07-15
fix(matchprop): match text style and height between TEXT/MTEXT objects

### LINKED ISSUE #361: Matchprop do not work on Text-Text and Text-Mtext
Matchprop do not work on Text-Text and Text-Mtext.

KevinGriffin-new:
Fixes #361

## Problem

MATCHPROP copied only the generic properties (layer, color, linetype, linetype scale, lineweight). Matching text to text left the destination's **text style and height** unchanged, so Text→Text and Text→MText matching appeared to do nothing.

## Fix

When the source object is a `Text` or `MText`, the `CmdResult::MatchProperties` handler now also copies `style` and `height` to any `Text`/`MText` destinations (all four combinations, both directions) — mirroring AutoCAD's "Text" special property. Non-text sources and destinations behave exactly as before. The existing `bump_geometry()` call re-tessellates, so matched text repaints at its new size immediately.

## Verification

- New in-process regression test `matchprop_matches_text_style_and_height` drives the real interactive command path (phase-1 source pick → phase-2 destination selection) with a Text source (height 5, style "BIG") and Text + MText destinations.
- On the unfixed code the test fails with the reported symptom (destination stays `STANDARD` / old height); with the fix it passes.
- Full suite: 203 passed, 0 failed.

### DIFF
--- a/src/app/automation.rs
+++ b/src/app/automation.rs
@@ -542,6 +542,69 @@ mod tests {
     }
 
     #[test]
+    fn matchprop_matches_text_style_and_height() {
+        // MATCHPROP between text objects must carry the text-specific
+        // properties (style, height) to TEXT and MTEXT destinations, not just
+        // the generic layer/color/linetype set. Regression for #361.
+        use crate::command::StepInput;
+        use acadrust::{EntityType, MText, Text};
+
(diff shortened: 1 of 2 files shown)


HakanSeven12:
thank you

PR #390, opened 2026-07-15
fix(export): render SDF text as vectors in PDF/print export (#385)

### LINKED ISSUE #385: PDF Export - does not print dimensions
Dimensions are not printing in current version.

Exporting in V.0.7.6 seems to work.
---
- Open CAD Studio: v0.8.2
- Platform: windows x86_64

KevinGriffin-new:
## Problem

Fixes #385. Since text became SDF-only (the legacy stroke-text path was removed in `5c15091`, made default in `9618156`), all text — standalone TEXT/MTEXT **and** dimension text — renders on-screen only as GPU glyph quads carried on `WireModel::text_verts`. The CPU PDF/print exporter (`build_pdf`) draws only wire stroke `points` and hatch fills, so it silently dropped every glyph. Dimensions and text were missing from exported PDFs and prints, while still displaying correctly on screen. This is the regression the reporters saw between v0.7.6 (good) and v0.8.2 (bad).

## Fix

Reconstruct vector text in the exporter rather than resurrecting the stroke path, so the SDF render path and its caching stay untouched:

- **`sdf_atlas`**: `GlyphAtlas::export_table()` snapshots each baked glyph's vector geometry (outline strokes + fill triangles, in glyph space) keyed by its atlas tile `uv_min`, re-resolved through `Face` so it honours the current `TEXTFILL`.
- **`pdf_export`**: a new `emit_text` pass walks each wire's `text_verts` (one 6-vertex quad per glyph), recovers the glyph geometry by `uv_min`, and affine-maps it into the quad — `DrawLine` polylines for stroke (LFF/SHX) fonts, filled triangles for TrueType, plus solid decoration bars (underline/overline/strike). It runs under the same rotation/scale/clip CTM as the wires, so it covers model space, paper space and windowed plots, and every SDF-text kind (text, dimension, multileader, table, tolerance). PDF export and printing share the same `export_pdf` entry, so both are fixed.

No changes to the hot, cached wire-building path — the SDF perf win is fully preserved.

## Scope note

Text **embedded in complex linetypes** (e.g. `GAS——GAS`) is SDF-only with no stroke/fill fallback in the glyph source, so it remains out of scope for this change.

## Tests

- Unit test: a stroke-font glyph emits vector draw ops, and every emitted point lands within the glyph's world bounding box (guards the coordinate mapping).
- Integration test (`tests/pdf_export_text_check.rs`): a real `Text` entity + linear `Dimension` flow through `scene.entity_wires()` → `export_pdf`; asserts text is carried to the exporter and grows the PDF vs. the same wires with text stripped.
- Verified **bug-first**: with `emit_text` disabled the end-to-end test fails with the reported symptom (text adds no content); restoring the call turns it green. `cargo test --lib` passes.

### DIFF
--- a/src/io/pdf_export.rs
+++ b/src/io/pdf_export.rs
@@ -322,6 +322,13 @@ fn build_pdf(
         flush_line(&mut ops, &segment);
     }
 
+    // Text (SDF glyph quads) — re-emitted as vector strokes / fills. Text now
+    // renders on-screen only as textured SDF quads (`wire.text_verts`), which
+    // this CPU exporter can't sample, so without this pass all text — including
+    // dimension text — is missing from the PDF (issue #385). Drawn after the
+    // wires (on top) and under the same rotation/scale/clip CTM.
+    emit_text(&mut ops, wires, ox, oy);
+
     if needs_state {
         ops.push(Op::RestoreGraphicsState);
     }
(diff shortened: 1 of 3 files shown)


HakanSeven12:
thank you

PR #392, opened 2026-07-15
fix(windows): show the app icon on the taskbar button (set ICON_BIG)

KevinGriffin-new:
## Problem

On Windows the taskbar button shows a blank/generic icon the whole time OCS runs, even though the title-bar icon is correct and the `.exe` embeds a proper icon.

Cause: iced's cross-platform `window::Settings.icon` maps to winit's `window_icon`, which on Windows sets only **`ICON_SMALL`** — the title-bar / Alt-Tab glyph. The taskbar button renders **`ICON_BIG`**, and iced/winit never set it. Confirmed live with `SendMessageW(hwnd, WM_GETICON, ICON_BIG)` returning `0` on a freshly opened window; sending `WM_SETICON(ICON_BIG)` populates the taskbar button immediately.

This affects every Windows install (reported on macOS too, though the Dock is a separate `.icns` surface — out of scope here).

## Fix

New Windows-only `src/win_taskbar.rs`, called from `main.rs` right before `app::run()`:

- Builds an `HICON` from the **same brand logo iced uses for the title bar** (`build_window_icon`), via `CreateIconIndirect` over a top-down 32bpp DIB with a zeroed AND mask — so title bar and taskbar always match, on dev and release alike.
- On a short-lived background thread (iced opens the window asynchronously after `app::run()` takes the main thread), it finds this process's titled top-level window and applies `WM_SETICON(ICON_BIG)` + `WM_SETICON(ICON_SMALL)`, and sets the class icon (`GCLP_HICON`/`GCLP_HICONSM`) so any later window inherits it.

No new dependencies — `windows-sys` was already a dependency; this adds its `Win32_Graphics_Gdi` feature.

## Verification

- **End-to-end on a release build**: launched the app and probed the live window — `ICON_BIG` went from `0` (stock) to the brand `HICON`.
- Unit tests: `rgba_builds_a_valid_hicon` (the RGBA→HICON path returns a real handle) and `rejects_undersized_buffer`. Full `cargo test --lib` green.

Follow-up idea (not in this PR): a higher-resolution taskbar icon by rasterizing `assets/logo.svg` at 256px rather than reusing the 32px title-bar bitmap.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,7 +86,7 @@ lyon_tessellation = "1.0.20"
 #                       (IApplicationAssociationRegistrationUI dialog).
 # Win32_System_Registry — HKCU registration that lists the app under
 #                       "Open with" for the portable .exe.
-windows-sys = { version = "0.61", features = ["Win32_UI_Shell", "Win32_UI_WindowsAndMessaging", "Win32_System_Com", "Win32_System_Registry", "Win32_Foundation"] }
+windows-sys = { version = "0.61", features = ["Win32_UI_Shell", "Win32_UI_WindowsAndMessaging", "Win32_System_Com", "Win32_System_Registry", "Win32_Foundation", "Win32_Graphics_Gdi"] }
 
 [patch.crates-io]
 # The `acadrust` crate now tracks the OpenAEC-Foundation/acadifc fork's main
--- a/src/app/helpers.rs
+++ b/src/app/helpers.rs
@@ -367,7 +367,7 @@ pub(super) fn title_case_word(value: &str) -> String {
 
 /// Builds a 32×32 RGBA icon: red background with OCS drawn in white pixels.
 #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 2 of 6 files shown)


HakanSeven12:
I cant reproduce this on windows. I tried both installer and portable versions.

KevinGriffin-new:
You're right, and thanks for testing both artifacts. I've traced it: my repro was on a local build, where `packaging/windows/AppIcon.ico` doesn't exist (it's generated by the release workflow from `assets/logo.svg`), so `build.rs` skips embedding the exe icon and the taskbar's fallback chain bottoms out at the generic icon. On installer/portable builds the exe carries the embedded icon and the taskbar resolves it fine — `ICON_BIG` being unset on the window doesn't matter because Explorer falls back to the exe's icon resource. So this "fix" only affects dev builds, and on release builds it would actually replace the crisp 256px embedded icon with a 32px one in large-icon surfaces (Alt-Tab previews, large-taskbar modes). Closing.

PR #586, opened 2026-07-30
feat(snap): Snap Store packaging

Nozzit:
## Summary
Adds `snap/snapcraft.yaml` + `.github/workflows/snap.yml` to bring Open CAD Studio onto the Snap Store, following the same repackage-the-release-artifact pattern already shipped for the other OpenAEC Foundation apps (open-planner-studio, open-calc-studio, open-geotechniek-studio, openaec-installer) — except this repackages the released **AppImage** (`--appimage-extract`) instead of a `.deb`, since that's what this project's `Release` workflow produces for Linux.

- Snap name: `open-cad-studio`.
- Triggers on completion of the **Release** workflow (`workflow_run`) — matches its `release: published` + `workflow_dispatch` triggers, so there's no risk of firing on routine pushes.
- Publishing to the Snap Store only happens once `SNAPCRAFT_STORE_CREDENTIALS` is set as a repo secret — until then the publish step is skipped, not failed.

## ⚠️ This one is meaningfully different from the other OpenAEC snaps — please read before merging

OCS uses **winit + wgpu** (Vulkan with a GL fallback) for its window/rendering, not GTK/WebKit. That means:
- No `extensions: [gnome]` (that's specific to the WebView-based Tauri apps). Instead this declares explicit `opengl`, `wayland`, `x11`, `desktop`, `desktop-legacy` plugs and stages the Mesa/X11/Wayland runtime libraries the binary needs (`ldd` on the released binary shows almost no linked libs beyond libc, which strongly suggests wgpu/winit load Vulkan/GL/X11/Wayland dynamically at runtime rather than at link time — the staged packages should cover that, but this is inferred, not confirmed against a real GPU).
- I have high confidence the **packaging mechanics** work (verified: extracting the actual v0.9.0 AppImage locally produces exactly the file layout the manifest expects, and `snapcraft.yaml` validates). I have much lower confidence in the **actual GPU/windowing behavior** under `strict` confinement — that can only really be verified by installing the built snap on a real Linux desktop with a GPU and confirming the app actually renders and opens/saves DWG/DXF files, not just that `snapcraft pack` succeeds in CI.

## Prerequisites before this can publish anything
1. The name `open-cad-studio` isn't registered on snapcraft.io yet.
2. `SNAPCRAFT_STORE_CREDENTIALS` needs to be added as a repo secret (`snapcraft export-login --snaps=open-cad-studio --channels=stable`).

## Test plan
- [ ] Merge, then run a `workflow_dispatch` test build against an existing tag (e.g. `v0.9.0`) — build/artifact only, no publish.
- [ ] Install the resulting snap on a real Linux desktop and actually exercise it: open a DWG/DXF file, confirm 2D/3D rendering works, not just that the window opens.
- [ ] If rendering is broken under strict confinement, this may need additional plugs/interfaces (or, in the worst case, `confinement: classic`, which requires manual Snap Store review on every release).
- [ ] Register the name and add credentials once the above checks out, then let the next real `Release` publish it.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -1,0 +1,169 @@
+name: Build Snap Package
+
+on:
+  # workflow_run werkt alleen als dit workflowbestand op de DEFAULT branch staat.
+  # De workflow wordt getriggerd zodra "Release" afrondt (die triggert zelf op
+  # `release: published` + workflow_dispatch, niet op elke main-push).
+  workflow_run:
+    workflows: ["Release"]
+    types: [completed]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Bestaande release-tag om tegen te bouwen (test)'
+        required: true
+        default: 'v0.9.0'
+
+# Minimale default; de build-snap-job zet zelf `contents: write` omdat hij de
+# snap als asset aan de release hangt.
+permissions:
+  contents: read
+
+# Eén globale groep, NIET annuleren: twee releases kort na elkaar publiceren
+# allebei naar het `stable`-kanaal van de Snap Store, en een Snap-release is
+# niet in te trekken.
+concurrency:
+  group: snap
+  cancel-in-progress: false
+
+jobs:
+  build-snap:
+    timeout-minutes: 75
+    runs-on: ubuntu-22.04
+    # `types: [completed]` vuurt OOK als de release-workflow faalde of werd
+    # geannuleerd; zonder deze gate zou een mislukte release alsnog een snap
+    # bouwen. Geen extra head_branch-guard nodig zoals bij open-calc-studio:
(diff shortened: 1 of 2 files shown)


UserDevtec:
Hi @Nozzit, 
Could you translate the notes and text into English instead of Dutch?

image

Nozzit:
I am going on vacation tomorrow, I'll do it later so don't worry when it takes a while

Nozzit:
Done — all inline comments and log/error strings are now in English (9a48ab6). No functional changes in that commit, comments and messages only.

Thanks for catching it; those comments were carried over verbatim from the sibling OpenAEC repos, where the working language is Dutch.

*(This comment was written and posted by an AI assistant working on @Nozzit's behalf.)*

PR #687, opened 2026-08-05
feat: add docked block palette

KarimJerbi:
Add a searchable block palette with previews, file import, collision-safe nested definitions, and dependency/table-record import. Keep palette thumbnails current across drawing switches and block-definition edits.


HakanSeven12:
That looks really nice :) thank you

PR #689, opened 2026-08-06
Fix LTSCALE refresh and circle linetype rendering

### LINKED ISSUE #667: BUG: LTSCALE is not working
This command is not making global changes to linetype scales.

Image

Image

gianlucafiore:
## Summary

This PR fixes two related linetype rendering issues:

- Global `LTSCALE` changes now immediately refresh existing drawing geometry.
- Circle entities now preserve and display their assigned linetype patterns.

## Problem

Changing `LTSCALE` correctly updated the document header, but existing entities continued displaying their cached wire geometry. The new scale only became visible after individually modifying an entity.

Circle entities also ignored non-continuous linetypes because the `TruckObject::Lines` tessellation path created their wire geometry without preserving the resolved linetype pattern.

## Changes

- Invalidate cached drawing geometry after changing global `LTSCALE`.
- Preserve `pattern_length` and `pattern` when tessellating circle entities.
- Keep the existing continuous behavior for other entities using the `TruckObject::Lines` path.

## Testing

Manually tested with:

- Lines
- Polylines
- Arcs
- Circles
- Continuous, dashed and dash-dot linetypes
- Multiple global `LTSCALE` values
- Existing entities created before changing `LTSCALE`

All existing entities now refresh immediately after changing `LTSCALE`, and circles correctly display their assigned linetype.

Fixes #667

### DIFF
--- a/src/app/commands/styleprops.rs
+++ b/src/app/commands/styleprops.rs
@@ -1887,6 +1887,8 @@ impl OpenCADStudio {
                     if v > 0.0 {
                         self.push_undo_snapshot(i, "LTSCALE");
                         self.tabs[i].scene.document.header.linetype_scale = v;
+                        // LTSCALE affects the cached linetype geometry of the whole drawing.
+                        self.tabs[i].scene.bump_geometry();
                         self.tabs[i].dirty = true;
                         self.command_line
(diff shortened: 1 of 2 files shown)


HakanSeven12:
thank you

PR #696, opened 2026-08-07
Add perpendicular object snap tracking

### LINKED ISSUE #695: Improvement: Perpendicular reference tracking
I'm going to address an improvement so that when drawing with the perpendicular snap on, it references the 90-degree recoil with respect to some element but taking that element as the origin, and not coming from outside.

Image

Currently, the perpendicular snap only allows referencing from outside. 

Image

gianlucafiore:
## Summary

Adds perpendicular tracking support for acquired object snap references.

Previously, Perpendicular OSNAP only worked from the current drawing point toward a target object. It could find the perpendicular foot on that object, but it could not use a point on an existing segment as the origin of a perpendicular tracking direction.

## Changes

- Store perpendicular tracking directions separately from regular edge tracking directions.
- Allow acquired points on line and polyline segments to generate a perpendicular OTRACK ray.
- Keep synthetic perpendicular rays separate from Extension snap geometry.
- Preserve the existing Perpendicular snap behavior from an external drawing point to a target object.

## Behavior

With Perpendicular OSNAP and OTRACK enabled:

- acquire a point on an existing line or polyline segment;
- move the cursor away from that point;
- the cursor can now lock to a 90-degree tracking ray relative to the referenced segment;
- tracking works in either direction from the acquired point.

## Testing

Manually tested with:

- Lines
- Polylines
- Endpoint references
- Intermediate points on segments
- Both sides of the perpendicular direction
- OTRACK enabled and disabled
- Extension snap enabled
- Existing Perpendicular snap behavior

Fixes #695

TEST

https://github.com/user-attachments/assets/c1d566a0-6db4-4d0d-bbdd-3d5e564c8424

### DIFF
--- a/src/snap.rs
+++ b/src/snap.rs
@@ -124,6 +124,12 @@ pub struct Snapper {
     /// then locks to that line (#219). Empty for a point that is not a segment
     /// endpoint (e.g. a midpoint or centre acquisition).
     pub tracking_dirs: Vec<Vec<DVec3>>,
+    /// Perpendicular directions through each acquired tracking point.
+    ///
+    /// Kept separate from `tracking_dirs` because those directions are also
+    /// consumed by Extension snap and extended-intersection logic. These rays
+    /// belong only to OTRACK when Perpendicular OSNAP is enabled. (#695)
+    pub tracking_perp_dirs: Vec<Vec<DVec3>>,
     /// Last snap world position (for dwell detection).
     pub last_snap_world: Option<DVec3>,
     /// When the cursor first rested near `last_snap_world`.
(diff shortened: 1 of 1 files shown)


semhustej:
Hi,
I tested the new feature in the latest release. It is a useful feature, thanks for the addition.

What I don't like about the current implementation is that it is too easy to create `Perpendicular` tracking points in random places on various objects. This leads to user being overwhelmed by temporary tracking points. See example in the video:


https://github.com/user-attachments/assets/5c35682c-02ec-4265-8cf5-040ce3667ba0



I think there should be only one perpendicular temporary tracking point per object or these tracking points should be only possible to create in other important points, like Endpoint or Midpoint. I am not sure how difficult this is to apply. The current behavior is not a deal breaker.

gianlucafiore:
Thanks! Yes, I agree, I've found it a bit overwhelming. Somewhat arbitrary reference points are added as the cursor moves, creating a complex nesting structure.

PR #711, opened 2026-08-08
Reject unsupported export extensions

DJ8520:
## What changed

- Validate native save/export target extensions before serialization.
- Reject missing or unsupported extensions with an error that names the supported `.dwg` and `.dxf` formats.
- Add a unit regression proving `.pdf` fails without creating a mislabeled output file.

## Why

`--export input.dxf output.pdf` previously fell through to the default DWG writer, exited successfully, and produced DWG bytes under a `.pdf` filename.

## Impact

Headless callers now receive a non-zero exit and actionable format guidance instead of silently receiving corrupt-by-contract output. Existing DWG and DXF save paths are unchanged.

## Validation

- `cargo test --release --locked --lib io::save_failure_tests::unsupported_output_extension_is_rejected_before_write -- --exact --nocapture`
- Production release build followed by a real `--export ... out.pdf` probe: exit 1, no output file, supported formats named.
- Existing `io::layer_roundtrip_tests` suite: 4 passed.

### DIFF
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -1336,6 +1336,27 @@ impl std::fmt::Display for SaveFailure {
 
 impl std::error::Error for SaveFailure {}
 
+const SUPPORTED_SAVE_FORMATS: &str = ".dwg, .dxf";
+
+fn validate_save_extension(path: &Path) -> Result<(), SaveFailure> {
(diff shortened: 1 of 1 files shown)


HakanSeven12:
Let me know when you finish it

PR #717, opened 2026-08-09
Fix perpendicular and OTRACK snap interactions

### LINKED ISSUE #716: BUG: Arbitrary reference points and perpendiculars
Hi,
I tested the new feature in the latest release. It is a useful feature, thanks for the addition.

What I don't like about the current implementation is that it is too easy to create `Perpendicular` tracking points in random places on various objects. This leads to user being overwhelmed by temporary tracking points. See example in the video:

https://github.com/user-attachments/assets/5c35682c-02ec-4265-8cf5-040ce3667ba0

I think there should be only one perpendicular temporary tracking point per object or these tracking points should be only possible to create in other important points, like Endpoint or Midpoint. I am not sure how difficult this is to apply. The current behavior is not a deal breaker.

_Originally posted by @semhustej in https://github.com/HakanSeven12/OpenCADStudio/issues/696#issuecomment-5226722424_
            

I'll be resolving it today.

gianlucafiore:
## Summary

Fix several related Perpendicular, Extension and Object Snap Tracking interactions.

- Prevent Perpendicular snap from behaving like Nearest when a command has no drawing base yet.
- Restrict automatic OTRACK acquisition to stable geometric snap points instead of continuous/contextual snaps.
- Allow OTRACK to resolve intersections while Extension snap is active, so Extension and tracking rays can meet at an exact point.

Fixes #716

### Perpendicular without a base point

Perpendicular previously fell back to the cursor position when no drawing base existed, such as while LINE was asking for its first point.

That effectively made Perpendicular behave like Nearest along nearby geometry.

Perpendicular now only produces a candidate when a real drawing base is available.

### OTRACK point acquisition

Automatic tracking-point acquisition now accepts stable geometric snaps such as:

- Endpoint
- Midpoint
- Center
- Node
- Quadrant
- Intersection
- Insertion
- Apparent Intersection

Continuous/contextual snaps such as Nearest, Perpendicular and Extension no longer leave arbitrary acquired tracking points while hovering.

Normal OSNAP hover behavior is unchanged.

### Extension + OTRACK intersections

OTRACK was previously skipped whenever Extension had already produced an OSNAP result.

OTRACK is now also evaluated while Extension is active, allowing the existing tracking-ray intersection logic to resolve crossings such as:

- Extension × Perpendicular tracking
- Extension × other OTRACK rays

The same behavior is applied during both cursor preview and point commit so the committed point matches the displayed tracking intersection.

## Testing

- Verified Perpendicular does not snap along geometry while LINE is requesting its first point.
- Verified Perpendicular works normally after a drawing base has been established.
- Verified stable OSNAP points can still be acquired by OTRACK through hover.
- Verified continuous/contextual snaps no longer create arbitrary tracking-point markers.
- Verified Extension continues to work from acquired endpoints.
- Verified Extension and Perpendicular OTRACK rays lock to their exact intersection.
- Verified the clicked/committed point matches the preview intersection.
- `cargo check --bin OpenCADStudio -j3` passes.
- `git diff --check` passes.


gianlucafiore:
Intersections with extensions can now be found
imagen


The issue with arbitrary anchor points has been fixed, as well as the permanent detection of perpendiculars.
imagen

With all these changes made over the last 3 PRs, I believe we have achieved maximum professionalism in the handling of references, snaps, and drawing assistance.

PR #721, opened 2026-08-10
Add REDRAW / REDRAWALL / REGEN / REGENALL commands

KarimJerbi:
## Summary

Implements `REDRAW`, `REDRAWALL`, `REGEN`, `REGENALL`, plus `R`/`RA`/`RE`/`REA` aliases.

- **REDRAW** - force a one-shot **re-rasterization** of the active viewport (geometry-pass invalidation only). Does not touch the document, `geometry_epoch`, `block_epoch`, undo, or the tab's `dirty` flag.
- **REDRAWALL** - same, but for every currently-generated viewport (all model tiles / sheet + paper content viewports).
- **REGEN** / **REGENALL** - full model regeneration via `Scene::bump_geometry()`. Functionally identical (all viewports share one `Scene`). Bumps `geometry_epoch` + `block_epoch` and clears tessellation caches, but never mutates document entities and never sets `dirty`.

## Architecture

The key problem: REDRAW's target must survive multi-pane Model layouts where the renderer runs `build_viewport_for_pane` **once per tile** per frame (and paper layouts run `build_viewports` once). A single shared "consume once" flag would let tile 0 eat the request before the others see it.

**Solution - per-instance-id refresh set:**
1. `Scene.request_refresh(scope)` resolves the scope (`Active` | `All`) to a concrete set of `instance_id`s at command time and stores it in `pending_refresh: RefCell<FxHashSet<u64>>`. `All` is a superset that dominates; `Active` falls back to the paper sheet when nothing is entered (D6).
2. Each builder drains **only its own id** via `refresh_consume(instance_id_for(inst))` → one-shot per viewport, no cross-tile race.
3. A consumed id lands on `ViewportData.force_rasterize`, which `prepare()` uses to pin `render_sig = u64::MAX` once → `skip_geometry = false` for that one frame → full geometry pass runs.

`instance_id_for` mirrors the renderer's existing id encoding (0x1… tile / 0x2… sheet / 0x3… paper-content), so resolution and consumption can never disagree.

## Design semantics (why the behavior is subtle but correct)

- A forced re-rasterization is **invisible** when the cached frame is already correct - that's expected and matches AutoCAD. It only matters when the display is stale. The `PERF` panel's `tess ms` / `epoch` readout makes it observable: REDRAW leaves `epoch` unchanged, REGEN increments it.
- REDRAW intentionally invalidates only the geometry-pass cache (`render_sig`), not the per-buffer GPU caches - a cheap, deliberately-scoped operation.
- REDRAW is **not** transparent and does not itself clear previews / cancel the active command; normal non-transparent dispatch teardown governs that separately.

## Files changed

- `src/scene/mod.rs` - `ViewportRefreshScope`, `pending_refresh` field, `instance_id_for` / `request_refresh` / `refresh_consume` (+ `#[cfg(test)] refresh_pending_any`), tests.
- `src/scene/view/render.rs` - `ViewportData.force_rasterize`, threaded through `build_viewports` + `build_viewport_for_pane` + `viewport_data_for`, consumed in `prepare`.
- `src/app/commands/display.rs`  - the four dispatch arms + tests.
- `src/app/commands/mod.rs` -  autocomplete registry entries.
- `assets/ocad.pgp` - `R` / `RA` / `RE` /`REA` aliases.

## Tests

7 new tests, all passing:
- Scene: resolve-then-consume per id; All "resurrects" an id after Active fully drained; All targets every existing model tile (via `split_active_pane`).
- Commands: REDRAW leaves epochs untouched; REDRAWALL leaves a force request; REGEN regenerates but does not dirty the document; alias `R` routes like `REDRAW`.

```
cargo test --lib redraw_tests commands::display::tests
```


HakanSeven12:
Hi @KarimJerbi do you want to be a collaborator?

KarimJerbi:
I am not a professional developer, but i always enjoyedwriting code. Willing to help and collaborate tho 😁.
i do use AI often to write code, but i do manual testing and I manually review it.

HakanSeven12:
> i do use AI often to write code, but i do manual testing and I manually review it.

Me too :) Project getting bigger and I need help from others. So I think you and @gianlucafiore can handle issues and 2D workflow and I can work on render, kernel, 3D modelling and dwg/dxf compatibility. I will send you a collaboration request

HakanSeven12:
Also I sent you a mail

KarimJerbi:
I received the email

PR #745, opened 2026-08-12
Fix viewport and annotation scales for drawing units

gianlucafiore:
## Summary

This PR fixes viewport and annotation scales so they correctly respect the drawing units defined by `INSUNITS` / `DWGUNITS`.

Previously, standard scales such as `1:100` were effectively applied as raw paper/model ratios without accounting for the physical size represented by one drawing unit.

This worked correctly for drawings modeled in millimeters, but drawings modeled in meters were treated as if their model units were millimeters.

## Problem

For example, in a drawing using meters:

- Model-space length: `10`
- Viewport scale: `1:100`

The expected result on paper is:

- `10 m / 100 = 0.1 m = 100 mm`

Previously, the `1:100` scale was applied as a raw factor of `0.01`, causing a 10-meter object to appear extremely small in paper space.

The viewport scale needs to account for the relationship between drawing units and paper units.

For a metric drawing in meters:

- `1:100` raw scale factor = `0.01`
- meters to millimeters = `1000`
- effective viewport factor = `0.01 × 1000 = 10`

This produces the expected 100 mm paper length for a 10 m object.

### Unit-aware viewport scales

- Add conversion between drawing units and paper units.
- Metric drawings use millimeters as the paper unit.
- Imperial drawings use inches as the paper unit.
- Apply this conversion to standard viewport scale factors.

### Annotation scales

- Make annotation multipliers account for the drawing unit.
- Apply the corrected multiplier when rendering annotative content inside paper-space viewports.
- Keep annotation scale values synchronized when `DWGUNITS` performs an actual geometry conversion.

### Scale list

- Make scale-list viewport factors unit-aware.
- Make scale-list annotation multipliers unit-aware.
- Keep the stored named scale itself unchanged, e.g. `1:100` remains a `1 / 100` scale definition.

### Viewport Properties

The viewport `Standard scale` property previously had a separate code path that parsed a label such as `1:100` directly into `0.01`.

That path had no access to the drawing units and therefore bypassed the unit conversion.

The Properties panel now uses the same Scene-level viewport scale logic as the normal viewport scale selector.

A new `set_viewport_scale_named_for()` path allows the centralized scale logic to be applied to a specific viewport.

The old viewport-local scale-ratio parser is no longer needed and has been removed.

### Viewport annotation rendering

The viewport camera could already have the correct effective scale while annotative viewport content was still using the raw annotation-scale inverse factor.

The viewport annotation multiplier now also accounts for the drawing-unit-to-paper-unit conversion, keeping the rendered model and annotation context consistent.

## Result

Standard viewport scales now behave physically correctly regardless of whether model-space units represent millimeters, centimeters, meters, inches, feet, or another supported drawing unit.

For example, with drawing units set to meters:

- `10` model units = `10 m`
- `1:100` = `100 mm` on paper
- `1:50` = `200 mm` on paper
- `1:200` = `50 mm` on paper

Custom workaround scales are no longer required simply because a drawing was modeled in meters instead of millimeters.

## Testing

Verified manually with two different cases.

### New drawing

- Set drawing units to `Meters`.
- Drew a line with length `10`.
- Applied standard viewport scale `1:100`.
- Result measured `100 mm` / `10 cm` on paper.

### Existing drawing

Tested with an existing DWG that had originally been modeled numerically in meters while its drawing units were `Unitless`.

For example:

- `8` drawing units represented an intended physical length of `8 m`.

After changing `DWGUNITS` to `Meters`:

- geometry correctly remained at `8`, because a Unitless drawing has no physical conversion basis;
- the standard `1:100` viewport scale correctly treated the geometry as meters;
- the same drawing now works with the normal standard scales instead of previously required custom workaround scales.

### Annotative content

Annotative text continues to preserve its assigned paper-space size while using the corrected unit-aware viewport scale.

### Build

Verified with:

`cargo check --release --bin OpenCADStudio`

The build completes successfully.

### DIFF
--- a/src/app/commands/layers.rs
+++ b/src/app/commands/layers.rs
@@ -876,6 +876,28 @@ impl OpenCADStudio {
             }
             reframed += 1;
         }
+
+        // Keep the active model-space annotation scale in sync as well.
+        // annotation_scale is drawing/paper, while CANNOSCALEVALUE stores
+        // the reciprocal paper/drawing factor.
+        if self.tabs[i].scene.annotation_scale.abs() > 1.0e-12 {
+            self.tabs[i].scene.annotation_scale *= factor as f32;
+        }
+
+        if self.tabs[i]
+            .scene
+            .document
+            .header
+            .annotation_scale_value
+            .abs()
+            > 1.0e-12
+        {
+            self.tabs[i]
+                .scene
+                .document
+                .header
+                .annotation_scale_value /= factor;
+        }
         // Sizes the drawing keeps as settings rather than as geometry: dash
         // lengths, default heights and widths, the radii the fillet and chamfer
         // commands start from. They are all lengths in the unit that just
--- a/src/app/update/command.rs
+++ b/src/app/update/command.rs
@@ -1720,12 +1720,29 @@ pub(super) fn on_tab_close(&mut self, idx: usize) -> Task<Message> {
                 Task::none()
     }
 
-    pub(super) fn on_prop_geom_choice_changed(&mut self, field: &'static str, value: String) -> Task<Message> {
-                let i = self.active_tab;
-                let handles = self.property_target_handles(i);
-                if !handles.is_empty() {
-                    self.push_undo_snapshot(i, "CHPROP");
-                    if field == "vp_ucs_name" {
+    pub(super) fn on_prop_geom_choice_changed(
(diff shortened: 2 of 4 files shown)


gianlucafiore:
The annotative scales are working now!!! :D

PR #770, opened 2026-08-15
fix(layers): handle case-only and current layer renames

### LINKED ISSUE #764: Layer name cannot be changed only by case
Renaming a layer from e.g. `TEST` to `test` in the Layers panel does nothing.

Steps:
1. Create a layer named `TEST`.
2. Double-click its name and rename it to `test`.

The layer should be renamed, but it keeps the original capitalization.

### LINKED ISSUE #765: Current layer name does not update after rename
Renaming the current layer in the Layers panel does not update the current-layer control in the ribbon. It keeps showing the old/stale layer until the layer is selected again from the ribbon.

Steps:
1. Make a layer current.
2. Rename it in the Layers panel.
3. Close the panel.

The ribbon should immediately show the new layer name.

NgoQuocViet2001:
## Summary
- allow case-only layer renames in the case-insensitive layer table
- update the active layer and document header when the current layer is renamed
- add regression tests for both rename paths

Fixes #764.
Fixes #765.

## Testing
- `cargo test layer_rename_tests --lib`
- `cargo clippy --lib`
- `cargo build --release --bin OpenCADStudio`

### DIFF
--- a/src/app/update/command.rs
+++ b/src/app/update/command.rs
@@ -1103,7 +1103,8 @@ pub(super) fn on_tab_close(&mut self, idx: usize) -> Task<Message> {
                         .get(idx)
                         .map(|l| l.name.clone())
                         .unwrap_or_default();
-                    let case_only_rename = new_name.eq_ignore_ascii_case(&old_name);
+                    let old_key = old_name.to_uppercase();
+                    let case_only_rename = new_name.to_uppercase() == old_key;
                     if !new_name.is_empty()
                         && new_name != old_name
                         && (case_only_rename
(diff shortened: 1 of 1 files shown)


src/app/update/command.rs
```diff
@@ -1103,31 +1103,40 @@ pub(super) fn on_tab_close(&mut self, idx: usize) -> Task<Message> {
                         .get(idx)
                         .map(|l| l.name.clone())
                         .unwrap_or_default();
+                    let case_only_rename = new_name.eq_ignore_ascii_case(&old_name);
```
HakanSeven12: `eq_ignore_ascii_case` disagrees with the table's Unicode `to_uppercase` keys, so Unicode case-only renames are still rejected as duplicates. Compare the normalized table keys.

src/app/update/command.rs
```diff
@@ -1103,31 +1103,40 @@ pub(super) fn on_tab_close(&mut self, idx: usize) -> Task<Message> {
...
                             if !nl.handle.is_valid() {
                                 nl.handle = self.tabs[i].scene.document.allocate_handle();
                             }
                             let _ = self.tabs[i].scene.document.layers.add(nl);
-                        }
-                        self.tabs[i].scene.document.layers.remove(&old_name);
-                        for e in self.tabs[i].scene.document.entities_mut() {
-                            if e.as_entity().layer() == old_name {
-                                e.as_entity_mut().set_layer(new_name.clone());
+                            for e in self.tabs[i].scene.document.entities_mut() {
```
HakanSeven12, line 1127: Renaming entity layer references leaves `Scene::dependency_index_cache` keyed by the old name. Later color or visibility changes can miss these entities. Invalidate the dependency index after renaming.

src/app/update/command.rs
```diff
@@ -1103,31 +1103,40 @@ pub(super) fn on_tab_close(&mut self, idx: usize) -> Task<Message> {
...
+                                if e.as_entity().layer() == old_name {
```
HakanSeven12: Exact equality skips valid case-insensitive entity references. A non-case-only rename can leave them pointing to the removed name. Match using the table's normalization.

src/app/update/command.rs
```diff
@@ -1103,31 +1103,40 @@ pub(super) fn on_tab_close(&mut self, idx: usize) -> Task<Message> {
...
+                                    e.as_entity_mut().set_layer(new_name.clone());
+                                }
+                            }
+                            if self.tabs[i]
+                                .active_layer
+                                .eq_ignore_ascii_case(&old_name)
+                            {
+                                self.tabs[i].active_layer = new_name.clone();
+                                self.tabs[i].scene.document.header.current_layer_name =
```
HakanSeven12, line 1145: When a handleless current layer receives a new handle, `current_layer_handle` remains invalid. Preserve the allocated handle and update the header together with `current_layer_name`.

HakanSeven12 (commented):
Static review found four rename-state regressions.

PR #1280, opened 2026-09-14
Wire up the Properties panel's Parameters section

felixriestra:
## Summary
`PropValue::ParamRow`/`ParamAddRow`, the `PropParamInput`/`Commit`/`Delete`/`AddNew` messages, their `command_driver` handlers, and their render functions all already existed — but nothing constructed the "Parameters" section that would ever route data to them, so the whole inline add/rename/redefine/delete workflow was unreachable. This adds that section to the no-selection page in `refresh_properties`, plus a `PropValue::ParamsVisibilityToggle` header row for toggling whether constraint pills show their driven value/parameter name.

Users can now manage named parameters inline in the Properties panel (live resolved values, add/rename/redefine/delete) as an alternative to the separate PARAMETERS modal, which keeps working unchanged.

## Test plan
- `cargo test --lib` — full suite passes (pre-existing, unrelated font-fallback test excepted)
- New tests: `no_selection_page_shows_every_named_parameter`, `add_rename_redefine_and_delete_a_parameter_through_the_panel` — verified both fail without the fix and pass with it


HakanSeven12:
All PR features are now integrated on main in merge commit dfe8b77c, updated to the current Properties and parametric APIs. The original commit and authorship are preserved in the merge history, and the visibility control reuses the existing shared setting.

PR #1295, opened 2026-09-15
Python scripting plugin: roadmap + plugin-architecture.md fixes

felixriestra:
## Summary

- Adds `docs/python-scripting-roadmap.md`: a phased plan for embedded Python
  scripting as an external plugin (RustPython, wrapping the existing
  `ocs_plugin_api`/`HostApi` contract), followed by a best-effort AutoLISP →
  Python transpiler built on top of it. Planning doc only — no host changes.
- Phase 1.1 of that plan is scaffolded and de-risked out-of-tree (not part of
  this PR): `rustpython-vm = "0.5"` builds cleanly as a `cdylib` dependency on
  macOS (arm64), confirmed with a real smoke-test command that runs the
  embedded interpreter, not just links it. The doc is updated with those
  findings.
- Two accuracy fixes to `docs/plugin-architecture.md`, found while
  cross-checking the scaffold against it:
  - The `Cargo.toml` example's `[patch.crates-io]` pointed `acadrust` at a
    stale repo/branch. The real source, per `ocs_plugin_api`'s own
    `Cargo.toml`, is a pinned git rev that isn't published to crates.io at
    all, so the patch block was both wrong and a no-op — `ocs_plugin_api`
    already pins it for any plugin that depends on it. Removed the block and
    explained why it's unnecessary.
  - The inline `CadModule` example showed `ribbon_groups()` returning an
    owned `Vec<RibbonGroup>`; the real trait (confirmed by a clean build
    against the live crate) returns `&[RibbonGroup]` via a `OnceLock`, same
    as `docs/plugin-template/src/lib.rs`. Fixed the example to match.
- `docs/plugin-template/src/lib.rs`: `MANIFEST.version` now derives from
  `env!("CARGO_PKG_VERSION")` instead of a hardcoded literal, matching the
  pattern used by the live reference plugin
  ([`opencad-example-plugin`](https://github.com/HakanSeven12/opencad-example-plugin))
  so it can't drift from `Cargo.toml`'s own version field. Verified the
  template still builds standalone.

## Test plan

- [x] `cargo build` inside `docs/plugin-template/` (standalone crate) — clean
- [x] External `opencad-python` scaffold (outside this repo) built with
      `rustpython-vm` as a `cargo build --release` cdylib on macOS arm64 —
      clean, correct C-ABI exports, working RustPython smoke test


felixriestra:
Withdrawing this one — it bundles plugin-specific planning docs (docs/python-scripting-roadmap.md) with the host-side changes, and the plugin work has moved to its own separate repo where it belongs. The legitimate ocs_plugin_api/automation changes here may get re-proposed cleanly on their own later.

PR #1299, opened 2026-09-15
Add Windows SpaceMouse navigation through 3DxWare

CiberNin:
Windows users can navigate drawings with a SpaceMouse through the installed 3DxWare driver. The app loads NavLib at runtime, exports CAD commands for device buttons and radial menus, and offers Follow context, Pan only, Pan and zoom, and 3D navigation.

Pan only and Pan and zoom let users push or tilt the puck to pan continuously. Both use raw multi-axis reports and share the app's Pan speed and Reverse pan direction controls; driver axis remappings do not apply to these modes. Full 3D navigation uses NavLib and honors the driver's Lock Horizon setting, with pitch limited at the poles to prevent the view from flipping.

Controls are in Options → User Preferences and the existing status-bar menu. Device callbacks queue proposed camera changes for the UI thread to apply to the active drawing, model tile, or entered paper-space viewport. Navigation pauses when the drawing loses focus or a dialog opens, respects viewport locks, and discards input from an earlier drawing or camera state. Button actions share the keyboard dispatcher, so Undo can remove the last point of an unfinished polyline.

## Validation

- `cargo test --locked --offline --lib`: **1,074 passed** in the default suite.
- Explicit run of the 15 normally skipped tests: **14 passed**. The existing disabled `cli_chprop_undo_restores_previous_state` test failed with the same assertion in a build without the SpaceMouse changes.
- `cargo check --locked --offline --bin OpenCADStudio`: passed.
- Installed-driver smoke test: passed; NavLib connected, detected the plugged-in device, and read Lock Horizon and the motion model.
- Physical testing confirmed continuous push/tilt panning in all four directions.
- Automated tests cover planar modes, release and stale input, horizon and pole behavior, focus and context changes, active commands, viewport camera round trips and locks, action IDs, and preference migration. Only Windows has a device adapter; other platforms show an unavailable state.

## UI

SpaceMouse status-bar menu

SpaceMouse preferences with pan direction and speed

The menu image uses a synthetic test drawing. The preferences image shows the current controls over a blank drawing.

[Usage and architecture](https://github.com/CiberNin/OpenCADStudio/blob/6a63726effb5c4607e92e6601f6939d5c42f4a1a/docs/spacemouse.md).


CiberNin:
I'm a bit unconfident about the UI positions I chose here.

PR #1326, opened 2026-09-17
Fix analytical intersection snapping for multi-segment polylines

yztink:
### Summary of Changes
Fixes an issue where snapping to the intersection of a circle (or arc/ellipse) and a polyline with two or more segments failed or snapped inaccurately.

### Root Cause
1. `curve_in_frame` and `wire_plane` in `src/snap.rs` strictly matched a single-element slice (`let [geom] = wire.tangent_geoms.as_slice() else { return None; }`). Any multi-segment polyline has 2 or more tangent geoms, causing `exact_curve_intersections` to return `None`.
2. As a result, intersection snapping silently fell back to discrete segment sweeping (`seg_intersect_3d`) against the circle's tessellated polygon chords (`WireModel.points`). Due to chord sagitta, the circle's straight chord cut across the corner between adjacent polyline segments or missed the corner vertex entirely.

### Fix
- Replaced single-slice matching in `wire_plane` with an iteration through `wire.tangent_geoms` to find any curved planar primitive.
- Generalized `curve_in_frame` to `curves_in_frame`, producing exact analytical curves for all polyline segments (both lines and bulge arcs) and utilizing `wire.key_vertices` for high-precision f64 endpoints.
- Intersected all segment pairs analytically via `cadkernel::geom2d::intersect` and deduplicated candidate points within numerical tolerance ($10^{-12}$) to cleanly handle shared corner vertices.
- Added unit tests verifying analytical intersections for multi-segment polylines against circles at shared vertices and with bulge arcs.


yztink:
first one, please go easy on me if it's royally messed up.... seemed to work great on my end...

src/snap.rs
```diff
@@ -2754,131 +2754,175 @@ impl WirePlane {
...
-        TangentGeom::Arc { center, axis_x, axis_y, radius, start_angle, end_angle } => {
-            let c = DVec3::new(center[0], center[1], center[2]);
-            let (ax, ay) = (DVec3::new(axis_x[0], axis_x[1], axis_x[2]), DVec3::new(axis_y[0], axis_y[1], axis_y[2]));
-            let n = ax.cross(ay).normalize();
-            if n.cross(frame.normal).length() > tol || !frame.contains(c, tol) {
-                return None;
+    for (geom_idx, geom) in wire.tangent_geoms.iter().enumerate() {
+        match geom {
+            TangentGeom::Line { p1, p2 } => {
+                let (p1, p2) = if wire.key_vertices.len() > geom_idx + 1 {
```
mechpanos, line 2820: This is the core problem. `key_vertices[geom_idx]`/`[geom_idx + 1]` only line up with the i-th `Line` geom for a plain LwPolyline with no bulges. Cases where they don't:

* **Polylines with bulge arcs** — `split_mixed_polyline` (`src/scene/convert/tessellate.rs:255`) emits every arc as its own wire and collapses the straight segments into one wire whose `tangent_geoms` contains *only* the `Line` geoms, while `key_vertices` keeps *all* polyline vertices. For `[L, A, L]` the straight wire has `tangent_geoms = [L0, L2]`, `key_vertices = [v0, v1, v2, v3]`, so `geom_idx = 1` reads `(v1, v2)` — the arc's chord — instead of `(v2, v3)`. This is the exact entity class the PR is meant to fix.
* **Center marks** (`src/entities/line.rs:20-36`) — `key_vertices` is two entries per segment plus the centre; `geom_idx = 1` reads `(end of seg 0, start of seg 1)`.
* **Block instances** (`src/scene/cache/block_cache.rs:2655-2676`) — several entities are merged into one entry; `key_vertices` are simply appended, and `transform_tangent` can drop a geom without dropping its vertices.

The tessellator guards against this at `tessellate.rs:411-425` by checking that `key_vertices[i]` matches `p1` (within 1e-3) before using it, falling back to `p1`/`p2` otherwise. Doing the same here (plus a `% key_vertices.len()` for the closing segment of a closed polyline, which currently falls through to the f32 branch) would fix all three cases.

src/snap.rs
```diff
@@ -2754,131 +2754,175 @@ impl WirePlane {
...
+
     if wire.tangent_geoms.is_empty() {
-        if wire.points.len() != 2 {
-            return None;
+        if wire.points.len() == 2 {
+            let (p0, p1) = (wp_f64(wire, 0), wp_f64(wire, 1));
+            if frame.contains(p0, tol) && frame.contains(p1, tol) {
+                curves.push(Curve::Line(KLine { start: frame.to_2d(p0), end: frame.to_2d(p1) }));
+            }
+        } else if wire.points.len() > 2 {
```
mechpanos, line 2800: This changes behaviour for wires with no `tangent_geoms` at all (splines, old-style 3D `Polyline`, hatch boundaries, etc.), which previously returned `None` here and went through the segment sweep.

With this branch every tessellation chord becomes a `Curve::Line`, the pair is marked resolved and the sweep is skipped. Three consequences:

* No precision gain — a chord is still a chord, so the result is the same as the sweep would give for the coplanar case.
* Cost: `points.len() × curves_b.len()` analytic intersections per wire pair, per snap frame, over every in-range wire pair. A dense spline near a few circles will feel this.
* Segments that aren't in the circle's plane are silently filtered out by `frame.contains`, so a 3D polyline that actually touches the circle is no longer reported, whereas `seg_intersect_3d` would have found it.

I'd keep the previous `points.len() != 2 → None` here so these wires stay on the sweep path.

src/snap.rs
```diff
@@ -2754,131 +2754,175 @@ impl WirePlane {
...
 
 fn exact_curve_intersections(wire_a: &WireModel, wire_b: &WireModel) -> Option<Vec<DVec3>> {
     let frame = wire_plane(wire_a).or_else(|| wire_plane(wire_b))?;
 
     const PLANE_TOL: f64 = 1e-7;
-    let curve_a = curve_in_frame(wire_a, &frame, PLANE_TOL)?;
-    let curve_b = curve_in_frame(wire_b, &frame, PLANE_TOL)?;
+    let curves_a = curves_in_frame(wire_a, &frame, PLANE_TOL);
+    let curves_b = curves_in_frame(wire_b, &frame, PLANE_TOL);
+    if curves_a.is_empty() || curves_b.is_empty() {
```
mechpanos, line 2901: Related to the alignment issue above: once `curves_in_frame` can return a *partial* or *wrong* set of curves for a wire, `Some(points)` here is no longer a safe signal for the callers to skip the sweep (`resolved_pairs.insert` at `snap.rs:1712`, `continue` at `snap.rs:1769`). Either the curve extraction has to be exact for every wire kind that reaches here, or this function needs a way to say "I handled the curved parts but the caller must still sweep". The first is simpler if the `Line` branch validates its vertices.

src/snap.rs
```diff
@@ -4171,4 +4215,68 @@ mod ext_tests {
...
+        assert!((ys[0] - 0.0).abs() < 1e-9, "expected corner intersection at y=0, got {}", ys[0]);
+        assert!((ys[1] - 10.0).abs() < 1e-9, "expected endpoint intersection at y=10, got {}", ys[1]);
+        for p in &pts {
+            assert!(p.x.abs() < 1e-9, "expected x=0, got {}", p.x);
+            assert!(p.z.abs() < 1e-9, "expected z=0, got {}", p.z);
+        }
+    }
+
+    #[test]
+    fn exact_curve_intersections_handles_polyline_with_bulge_arc_against_circle() {
```
mechpanos, line 4251: Both new tests build the `WireModel` by hand with `tangent_geoms`/`key_vertices` already aligned the way the new code expects, so they can't catch the `split_mixed_polyline` case. Suggest routing at least one test through `split_mixed_polyline` (it's in `src/scene/convert/tessellate.rs`, see the existing `test_split_mixed_polyline_*` tests for the call shape) or through a real `LwPolyline` entity with a bulge, and asserting on the *straight* wire it produces. That's the wire the snap engine actually receives for a polyline with arcs.

mechpanos (changes requested):
Thanks for digging into this — the diagnosis in the description is right (the `let [geom] = …` slice patterns did push every multi-segment polyline onto the chord sweep), and iterating all `tangent_geoms` is the correct direction. Existing snap tests and the two new ones pass locally.

However the `TangentGeom::Line` branch assumes `key_vertices[i]` / `key_vertices[i+1]` are the endpoints of the i-th tangent geom, and that invariant does not hold for several wire kinds the snap engine sees. Because `exact_curve_intersections` now returns `Some` for those wires, the callers at `snap.rs:1704` and `snap.rs:1764` mark the pair as resolved and **skip the segment sweep**, so the result is not "same as before" but lost intersections plus phantom ones. Details inline.

**Reproduction (real pipeline, not a hand-built `WireModel`)** — polyline `(-10,0)→(0,0)` straight, bulge arc `→(0,8)`, straight `→(10,8)`, pushed through `split_mixed_polyline` exactly as the tessellator does for any LwPolyline/Polyline2D with a bulge, then the straight-segment wire against a circle centred `(5,4)` r=5:

```
straight wire tangent_geoms=[Line(-10,0→0,0), Line(0,8→10,8)]
key_vertices=[(-10,0),(0,0),(0,8),(10,8)]
points = [DVec3(0.0, 4.0, 0.0)]
```

The two real crossings on `y=8` (x=2 and x=8) are missing and the single point returned lies on the arc's chord `x=0`, which is not part of this wire at all. Before this PR that pair fell through to the sweep and both real crossings were found.

What I'd suggest:

1. Don't rely on index alignment. Either validate `key_vertices[i]`/`[i+1]` against the geom's own `p1`/`p2` before using them (the tessellator already does exactly this at `tessellate.rs:411-425`, with a modulo for the closing segment), or just use the geom's `p1`/`p2` and address the f32 precision separately.
2. Keep returning `None` for wires with empty `tangent_geoms` and more than two points, so splines / 3D polylines continue to go through the sweep (see inline).
3. Add a test that goes through `split_mixed_polyline` (or a real `LwPolyline` entity with a bulge) rather than a hand-assembled `WireModel`, since the hand-built ones encode the very alignment assumption that fails in practice. A center-mark or block case would be good too.

Happy to re-review once that's in.

PR #1330, opened 2026-09-17
feat(ui): right-click context menus with command keywords, PLINE/ARC keyword parity, grip menu

mechpanos:
## Summary

Right-click in the drawing area now behaves the way drafters coming from commercial solutions expect, while keeping OpenCADStudio's own look and adding a few things they don't have.

### Context menu (`src/ui/popup/context_menu.rs`, `src/app/update/context_menu.rs`)
- The menu is built from a pure model (`build_context_menu`) that the overlay renders and the keyboard handler walks, so what is shown and what Enter picks can never diverge.
- **While a command runs**: Enter (default row, anchored under the pointer) · Cancel · every keyword of the command's `options()` with the keyword shown as a hint · Recent Input ▸ · Snap Overrides ▸ (every object snap with its glyph, Mid Between 2 Points, None, Osnap Settings…) · Pan / Zoom running transparently (`'PAN`, `'ZOOM …` — an interactive ZOOM Window parks the command and resumes it afterwards).
- **Idle**: Repeat, Recent Input ▸, Clipboard ▸, Undo/Redo with the history label, Isolate ▸, Pan/Zoom/Zoom Extents, Select All, Quick Select, ✓ Properties, Options.
- **With a selection**: Erase, Move, Copy Selection, Scale, Rotate, Mirror, Draw Order ▸, Select Similar, Invert Selection, Deselect All.
- **Hot grip**: Enter, Stretch, Move, Rotate, Scale, Mirror, Base Point, Copy (keeps the grip hot for repeated copies), Undo, Exit.
- Keyboard: ↑/↓, Enter, mnemonic letters (`C` = Close, …); any other key closes the menu and goes to the command line. Rows are 22 px, panels stay clear of the overlaid prompt lines (`CommandLine::overlay_lines_height`).
- Preference **Options → User Preferences → Right-click Customization**: Shortcut menu (default) / Time-sensitive (hold threshold slider) / Enter first (the previous behaviour). Also `SHORTCUTMENU` and `SHORTCUTMENUDURATION` sysvars.

### Command keyword coverage
- **LINE**: Close / Undo / eXit as `options()`.
- **PLINE**: Width, Halfwidth, Length; arc mode Angle, CEnter, CLose, Direction, Radius, Second pt with the nested prompts (Angle → CEnter/Radius, CEnter → Angle/Length, Radius → Angle). Per-segment widths land in the vertices; PLINEWID persists per session.
- **ARC**: the plain command now follows the compatible prompt flow — start [Center], second point [Center/End], end; after Center: end [Angle/chord Length]; after End: center [Angle/Direction/Radius]; Enter continues tangent. The former default stays as `ARC_CSE`.
- **MOVE** Displacement, **COPY** Displacement/mOde/Undo, **FILLET** Undo, **CHAMFER** Multiple/Undo, **OFFSET** Erase/Layer/Undo/Exit (new `CadCommand::on_document_undone` hook refreshes cached entities after an in-command undo). POLY3D, TRACE, MIRROR, MEASUREGEOM expose the keywords they already accepted.
- Object snaps gain a one-shot **None** override.

### Also
- Automation port: `pointer_right_press` / `pointer_right_release` actions.
- The pinned prompt keeps informational bracket groups ("[3 objects]") and drops only real option listings.
- All new strings in the 21 locales; catalogue rows added.

## Test plan
- `cargo test --locked --lib`: 1367 pass; the remaining failures on my machine are locale-dependent assertions (Greek UI) that fail on `main` too.
- New tests: context-menu model (14), app-level menu / modes / keyboard / grip (23), PLINE (17), ARC flow (6), MOVE/COPY/FILLET/CHAMFER/OFFSET options (9), transparent Pan/Zoom (5).
- Manual: LINE/PLINE/ARC/CIRCLE menus, selection and grip menus, the three right-click modes, `'ZOOM` inside LINE, Snap Overrides ▸ Endpoint/None on the next pick.


MarcelP102:
Hi @mechpanos 

Really nice addition! 

Some feedback and suggestions:
- Using ZOOM option shows the Hand as icon, perhaps use the icon of the menu? Also a selection box appears when holding left mouse button. (perhaps for later; but how cool would it be if the most used command show a different mouse. So you always know what command is on. For exmaple in SketchUp, the mouse icon changes to match your selected tool, see some examples: https://help.sketchup.com/en/sketchup/using-your-mouse Making the program a litlle bit more noob-friendly and less an hardcore-cad app?)

- Perhaps make a fly out menu instead of extending the menu downwards for sub-menu opening? Jumping menu items is a bit distracting.

- Copy with basepoint shows the coordinates making the menu unreadable. Perhaps this is a command error  because if I want to rerun this command I want it to ask again what the point should be. Show than it will show Repeat COPYBASE instead of Repeat COPYBASE xxxxxxx, xxxx. How does autocad handle this?

afbeelding

- Perhaps also add icons for other command like move/copy etc. We already have them in the ribbon so we can reuse them?

- to save some space perhaps not show the title of the command for the basis tools like move/copy etc? And add them like windows 11 right click menu does? That is also a combination of menu with only icons and traditional right click menu options
- 




-  Perhaps when nothing is selected right click also show the most used draw options (arc, circle, rectangle etc) (see also the QUAD of Bricscad below)

- Perhaps we could also add icons to the commandbar options you show? 

**Next-step?**
I really like that you show the commandbar options in the menu while running the command! What if we could extend this to make it even more context aware? I already gave the grip-menu (hovering over a control point) some love and added some command. What if you could make this looks the same as your menu or combine it? With icons for easier recognition?

Some examples:
- Bricscad has the QUAD, a context aware menu. https://help.bricsys.com/en-us/document/bricscad/getting-started-with-bricscad/exploring-the-interface/the-quad-basics?version=V26&id=165079147018
afbeelding
afbeelding


- Archicad has the pet pallet. Please also see this movie:  https://www.youtube.com/watch?v=MYQITM225GY
afbeelding 


That one is more icon driven, only icons and text only shown when hovering over the icon. I like that it start with the most recent used option of the menu. So for example when you click a control point of an arc and used the radius option before that will be used again instead of the stretch option. 

- Vectorworks has a floating menu. In the settings you can select what group  command are shown.  The tool modes are like what the options are in the command bar.
afbeelding

MarcelP102:
Hi @mechpanos 

Did some vibe concept designing. 
E058940E-EE11-496F-91A0-88448D68DA5C

mechpanos:
Great valuable comments Marcel! Will consider extending the capabitities, but needs more work and testing, for sure is more easy to add in the right menu functionality that is already existing.

PR #1353, opened 2026-09-18
feat(options): OK / Apply / Close with a discard guard

mechpanos:
Stacked on #1344 (base branch `plot/b3-psetupin`); only the last commit is new here.

The Options window used to commit every control the moment it was touched, with no way back but re-editing — easy to make a wrong choice and hard to undo. It now works like the Drafting Settings window:

- **OK** commits and closes; **Apply** commits (enabled only when something changed); **Close** — and the window's × or Esc — puts everything back the way it was when the window opened or when Apply was last pressed, asking first (*Unsaved changes will be discarded — Discard & close / Keep editing*) when there is something to lose.
- Changes still show at once, so a theme, a cursor size or a language is seen immediately and can be judged before committing.

The commit point is a snapshot of the persisted configuration plus the drawing variables the 3D Modeling page edits (ISOLINES, DISPSILH, SURFU/V/TYPE, SOLIDHIST, SHOWHIST, with the drawing's modified flag), so every control the window has or gains later is covered without being registered anywhere. Reverting re-applies the snapshot the way start-up does and re-bakes text (TEXTFILL) or regenerates solids (ISOLINES) when those moved.

The "unsaved changes" guard is now a shared helper (`ui::modal::discard_guard`) used by both windows.

Tests: live change + Close reverts, Apply becomes the new baseline, OK commits and closes, a clean Close asks nothing, × asks like Close, Keep editing returns with the change intact, Modeling-page drawing variables are reverted and the drawing's modified flag restored.


CiberNin:
Do we have a common convention where checkboxes = "takes effect on submission" and switches = "takes effect immediately"

Issue #1261, opened 2026-09-14
System printers


JCR-26:
OCS don't recognize all my system printers in my win 11 desktop and the default printer doesn't match with main. Only works by exporting to PDF. Not to mention the impossibility of manage legacy ctb

KarimJerbi:
legacy CTB is supported 

<img width="295" height="248" alt="Image" src="https://github.com/user-attachments/assets/ed7fbe6c-0740-42cb-9f55-f8840bde89db" />

<img width="791" height="553" alt="Image" src="https://github.com/user-attachments/assets/42873f7c-643c-4c5e-abd2-df9b42e23f9e" />


windows printer recognition issue is known, contributions are welcome !

JCR-26:
When I refer to legacy ctbs, I mean that despite manually copying my CTB files into the Plot Styles folder, OCS still shows them as 'missing' or 'not loaded' when I select them and try to run a plot preview. I think this is a known issue and only OCS originals ctb like ocad.ctb work properly

CiberNin:
Adding a reproducible Windows case and an enumeration probe.

On v2026.37 / Windows 11, Ctrl+P lists only **Default** and **Print to PDF**. Windows enumerates five printers on the same machine.

The Windows branch of [`list_printers()`](https://github.com/HakanSeven12/OpenCADStudio/blob/v2026.37/src/io/print_to_printer.rs#L95) returns an empty vector. The dropdown can still display a printer name stored in the drawing, which makes it look as though the printer was detected.

The attached probe uses `EnumPrintersW` with `LOCAL | CONNECTIONS`, level 4. It returned all five printers in about 5 ms and sent no print jobs. This could supply the missing Windows enumeration; named-printer dispatch still needs integration testing.

[01-printers.zip](https://github.com/user-attachments/files/32207598/01-printers.zip)

CiberNin:
Can OP verify if printer issue is resolved? Should CTB be split to a separate issue and this one closed?

KarimJerbi:
I can verify that the issue is solved on my machine, with brother printer. 

I think its worth keeping the issue until we investigate cbt import issues.

JCR-26:
[...]

PR #1358, opened 2026-09-18
fix(print): send Windows plots to the chosen printer, name the default, report failures

mechpanos:
Part of #1261 ("System printers"). The printer enumeration itself landed in `f85b23bf` (after the v2026.37 release); this PR fixes what was still wrong around it on Windows.

## Problems

- With no printer chosen, a plot went to the PDF application's bare `print` verb — whatever that application considered its default — and a viewer without a `printto` verb (Edge, store readers) silently dropped the printer the dialog had chosen, because a "successful" ShellExecute skipped the GDI path.
- The dialog never named the system default printer (`default_printer_name()` returned `None` on Windows).
- A failed enumeration left only "Default" and "PDF", with the reason written to a console that release builds do not have (`windows_subsystem = "windows"`).

## Changes

- **Default printer by name** (`GetDefaultPrinterW`): the dialog's entry reads "System default printer (NAME)", and every job resolves "Default" to that name before dispatch. Properties… on the Default entry opens that printer's preferences sheet.
- **Route decision** (`plan_print_route`, pure and unit-tested): the PDF application is used only when it registered a `printto` verb for `.pdf` (`AssocQueryStringW`), always with the explicit, quoted printer name — never the bare `print` verb; otherwise the job goes straight to the spooler through the existing GDI path. The report says which: `Sent to printer: NAME (direct print | via the PDF application)`.
- **Enumeration failures are visible**: `list_printers()` returns the system's own error text; the plot dialog shows "Could not list printers: …" under the printer row and the command line warns once. A desktop without CUPS is an empty list, not an error.
- **`PRINTERS` command**: default printer, listed printers (or the error) and, on Windows, the print route — a report to paste into an issue instead of a screenshot.
- GDI path: the driver's `dmSize` is kept (its private data lives behind it; it was being overwritten with the public struct size), the 300 DPI print raster is no longer memoised for the session (tens of MB per job), and a spooled temporary PDF is removed after a direct job.

## Verification

- Linux: `cargo test --lib` (route planning, `PRINTERS`, plot dialog tests) — only the failures already present on `main` remain.
- `cargo check --lib --tests --target x86_64-pc-windows-gnu` clean (the Windows-only code compiles; the two existing Windows GDI tests take the resolved name).
- Run-testing on Windows 11 with real printers is in progress (the maintainer's machine); the checklist: printer list, "System default printer (…)" naming, `PRINTERS` output, a plot to Default / a named printer with Edge vs a reader that registers `printto`, and the stopped-spooler error note.

No commercial product names in code or strings.


mechpanos:
**Windows testers wanted.** This branch was written on Linux and is compile-checked for the Windows target (`cargo check --lib --tests --target x86_64-pc-windows-gnu`), but nobody has run it on a Windows machine with real printers yet. If you have Windows 10/11 and a printer or two, a native `cargo build --release` of this branch and a run through the list below would settle it:

1. Ctrl+P: does the list show every printer from Settings → Printers & scanners? Does the *Default* entry read "System default printer (…)" with the same name Windows shows as default?
2. Type `PRINTERS` on the command line and paste its output here (default, list or error, print route).
3. Plot a layout to *Default*, then explicitly to a named printer: does each job land in that queue only, and what does the command line report — `Sent to printer: NAME (direct print)` or `(via the PDF application)`? If you can, try once with Edge as the `.pdf` handler and once with a reader that registers a `printto` verb (Acrobat).
4. A landscape and a portrait A4 on a physical printer: full size, correct orientation?
5. Stop the *Print Spooler* service and reopen the plot dialog: is there a "Could not list printers: …" note under the printer row? (Start it again afterwards.)
6. *Properties…* with *Default* selected: does the default printer's preferences sheet open?

Anything odd — the `PRINTERS` output plus a screenshot is enough.

PR #1359, opened 2026-09-18
feat(plot): Windows printer media and printable areas

mechpanos:
Part of #1261; **stacked on #1358** (base branch `plot/b4-windows-printer-routing`), only the last commit is new here.

The plot dialog already asks CUPS printers for their sheets and printable areas (#1311); on Windows that query was a stub, so every printer showed the paper catalogue with unverified margins and PAPERUPDATE had nothing to work with.

## Changes

- **`io/windows_media.rs`** (compiled and tested on every platform): the three parallel `DeviceCapabilities` tables — `DC_PAPERS` ids, `DC_PAPERSIZE` in tenths of a millimetre, `DC_PAPERNAMES` in 64-WCHAR slots — become one `PrinterMedia` per distinct size: a rotated or borderless twin folds onto its sheet (marking it borderless), catalogue sheets keep their own name, unknown sizes carry the driver's name, the device's margins go on every entry, and the driver's default sheet is found by id (`dmPaperSize`) or by user size (`DMPAPER_USER` + width/length).
- **Windows query** (`print_to_printer::windows_printer_media`): `DeviceCapabilitiesW` for the tables, the default `DEVMODE` for the default sheet (fetch shared with the print path through `with_printer_devmode`), one portrait printer DC for the unprintable margins (`PHYSICALOFFSET*`, `PHYSICAL*`, `HORZRES`/`VERTRES`, `LOGPIXELS*`) — per device, a documented approximation; a queue that lists no sheet (fax, virtual) keeps the catalogue. The query stays in the dialog's background task, as offline network queues can take seconds.
- **Direct prints select the plot's sheet** in the DEVMODE — by the driver's id when it lists that size, otherwise as a user-defined size — so an A3 plot prints on A3 rather than shrunk into the driver's default; a driver that refuses the sheet is asked for the orientation alone.
- The capability cache remembers printers that could not be asked, so a slow failing query runs once per session instead of on every dialog open.
- **`PRINTERS <name>`** reports a printer's sheets, margins and default sheet.

Nothing changes in the dialog itself: `PrinterMedia` answers, the "reported by the printer" sheet list, PAPERUPDATE and the paper-space printable-area guide now simply work on Windows too.

## Verification

- Linux: `cargo test --lib windows_media` (names, folding/dropping, default by id / user size / none, DC margins, paper-id lookup) plus the full suite.
- `cargo check --lib --tests --target x86_64-pc-windows-gnu` clean.
- Windows 11 run-testing in progress: select a printer → its sheets and margins appear; a sheet the printer lacks → PAPERUPDATE note; offline printer → dialog stays responsive; "Microsoft Print to PDF" with A3 selected → a 297 × 420 mm PDF; `PRINTERS <name>` output.


mechpanos:
**Windows testers wanted** (same caveat as #1358: written on Linux, compile-checked for the Windows target, not yet run on real hardware). With a native `cargo build --release` of this branch:

1. Select a printer in the plot dialog: does the paper list switch to the driver's sheets, with the "reported by the printer" note? `PRINTERS <printer name>` prints what was read (sheets, margins, default sheet) — please paste it.
2. In a paper layout, does the dashed printable area follow the printer's margins (compare with the driver's Printing preferences)?
3. Pick a sheet the printer does not have (e.g. A3 on an A4 printer), then switch printers: a PAPERUPDATE message should appear (and `PAPERUPDATE 1` switches to the driver's default sheet).
4. Print to *Microsoft Print to PDF* with A3 selected: is the produced PDF page 297 × 420 mm?
5. Select a printer that is offline / a network queue: the dialog should stay responsive and fall back to the catalogue.

PR #1360, opened 2026-09-18
fix(plot-style): load plot style tables written by other applications

mechpanos:
The CTB half of #1261 (independent of #1358 / #1359; branched from `main`).

Tables copied from other applications into the plot styles folder (`%APPDATA%\OpenCADStudio\plotstyles` on Windows) were listed in the Plot dialog but refused on selection, showing as "missing" — with the reason written to the command line only, and lost entirely when a page setup named the table. The loader failed hard on a checksum or text length that did not add up and on **any non-UTF-8 byte**; files from other applications carry their descriptions in Windows-1252, so a Spanish or Greek description was enough to reject the whole table.

## Changes

- **Lenient container decoding** (`decompress_ctb`): a checksum or length mismatch is reported and read past (a genuinely cut-off file is still an error), a zlib stream with a bad trailer is retried as raw deflate, and text that is not UTF-8 is read as Windows-1252 — the pens and colours are numbers and unaffected; a non-Western code page only shows wrong characters in descriptions.
- **`PlotStyleTable::load_warnings`**: every load site reports what had to be tolerated on the command line, so a table that needed tolerance says so instead of loading silently.
- **Errors that explain themselves**: a `.stb` name says named tables are not supported yet (station C1 of the plotting track); a `.ctb` missing from the folder names the folder to copy it into, instead of "Unknown built-in plot style".
- **The Plot dialog shows the reason** under the Table row — from a selection, from a page setup that names a table, and from the Load… picker, whose error used to be swallowed (`pick_plot_style` now returns it).
- A table picked with Load… from outside the plot styles folder is copied into it (unless a file of that name is already there), so it is listed from then on and page setups naming it resolve after a restart — what the "I copied my CTB files" workflow actually needs.

## Verification

- Tests build synthetic tables with `compress_ctb` and damage them: flipped checksum, text length off by one, a short payload field, Windows-1252 description bytes (`Plumas – señal`), a truncated stream (still an error), `.stb` and missing names.
- `cargo test --lib` — only the failures already present on `main` remain; two new strings in all 21 locales.
- Pending from the reporter: one of the actual CTB files that failed, to add as a fixture.


mechpanos:
**Testers wanted** (any platform): copy a `.ctb` written by another CAD application — ideally one with accented characters in its descriptions — into the plot styles folder (Windows: `%APPDATA%\OpenCADStudio\plotstyles`, Linux: `~/.config/OpenCADStudio/plotstyles`, macOS: `~/Library/Application Support/OpenCADStudio/plotstyles`) and select it in the plot dialog. It should load, with any tolerated defect reported on the command line, instead of showing as missing; a file that still fails should say why under the *Table* row. A copy of a file that still fails would make a good fixture.

PR #1383, opened 2026-09-19
docs(automation): consolidated vendor-neutral API specification

huaninstratech:
## What

Adds `docs/automation/API-SPEC.md`: a single, self-contained reference for external client developers of the automation API — conventions and units, the four transports, session/state, the full symbolic error catalogue with the retryable set, the complete operation catalogue (documents and identity, entity create/modify/query with `where` filters, block lifecycle, layouts and plotting, metadata and sysvars, groups and selection sets, embedded vs linked images, commands and batches), the REST endpoint table, worked examples and the versioning policy.

The document deliberately names only open formats and protocols (DWG/DXF/PDF, HTTP/JSON, OpenAPI, MCP) — no third-party products or vendors — so it can be handed to any client team as the API contract.


HakanSeven12:
Thanks for the write-up — it is a clear document, but I can't merge it as it stands: it describes an API surface that this repository does not have, so anyone reading `docs/automation/API-SPEC.md` here would try operations that do not exist.

What `main` actually exposes today (checked against `src/app/control/mod.rs` and `src/mcp.rs`):

- **Transports:** `--serve` (stdio JSONL), `--serve --port N` (TCP) and `--mcp`. There is no `--http` REST transport, no `/api/v1/…` routes and no OpenAPI document.
- **Operations:** `new`, `open`, `save`, `activate`, `select`, `run`, `start`, `input`, `cancel`, `stop`, `action`, `property`, `set_properties`, `embed_image`, `capture`, `undo`, `redo`; reads `state`, `capabilities`, `query` / `entities`, `records`, `record_schema`, `layers`, `header`, `properties`, `measure`, `history`.
- **Not present:** `entities_create`, `entities_transform`, `entities_delete`, `entities_copy_to`, `block_define` / `block_delete`, `wblock`, `layout_create`, `page_setup_set`, a `plot` op (and `per_page`), `sysvar`, `xdata_set` / `xdata_get`, `file_identity`, `group_create`, `selection_set_*`, `batch`, `events`, `view_focus`, `close`, the `where` filters, and an `ocs_execute` that accepts 35 mutations.

I take it the document describes the surface of your fork (the v2026.40 build mentioned in #1382). Two ways this can land:

1. Trim it to what upstream has today — the operations above, the three transports, the existing error codes — so the document stays true here; or
2. Upstream the operations first, in reviewable pieces, and bring the matching section of the spec along with each one.

Either is welcome; a spec that runs ahead of the code is not something I can keep in the tree. Two small things for whichever version comes back: the project is spelled `OpenCADStudio` (not "OpenCAD Studio"), and `docs/automation/README.md` already covers the MCP set-up, so the new file should link to it rather than restate it.

Issue #1318, opened 2026-09-16
Problem use


JPqcuenca:
panicked at src/modules/draw/modify/trim.rs:3982:42:
range start index 1 out of range for slice of length 0
panicked at /home/runner/.cargo/git/checkouts/winit-865c6d5afb3ec9c3/05b8ff1/src/platform_impl/web/event_loop/runner.rs:599:30:
RefCell already borrowed

---
- Open CAD Studio: v2026.37+1.g95bad2a3
- Revision: 95bad2a3b2e2 (2026-09-13, release)
- Platform: Web — Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/153.0.0.0 Safari/537.36

Issue #1313, opened 2026-09-16
OTRACK - when two vectors are used, only one of them is displayed


semhustej:
When using OTRACK to select intersection of two vectors (tracking from two points), when both vectors should be displayed, only one is visible.

The functionality works (snap to the intersection of two vectors), but it is not apparent to the user.

See comparison of OCS and BricsCAD (where it is clear, including the intersection snap symbol):

https://github.com/user-attachments/assets/3dc061dc-350c-46b4-926b-45b80c0fdcc3

```
Open CAD Studio v2026.37
Revision: a43fb253f64e
Commit date: 2026-09-13
Profile: release
Features: none
OS: Windows
Arch: x86-64
```

I am not sure, but I suspect this might be regression.

Issue #1259, opened 2026-09-14
Can not move block attributes


jprrezende:
When select blocks the softwares does not show the grip points of the atributtes. So I can not move atributtes to other position.

Thanks for attention and congrats for the software, this is the best open alternative to the AutoCAD.

PR #1386, opened 2026-09-19
Add some enhancements

dasmatus:


HakanSeven12:
Thanks — the three features in here are on `main` now, picked as their own commits with the original authorship:

- TRIM: skip polyline segments with no plan-view shape (#1318) — `e83c8a28`, `41101efa`
- OTRACK: draw both vectors of an intersection lock (#1313) — `033319a4`, `5d4b7072`
- INSERT: a grip per block attribute (#1259) — `9af327ed`, `bd89507a`, `4559c8a2`, `2b92fc5b`, plus `fd2062bd` for the two points Copilot raised: an attribute that ATTMODE 2 displays is now movable, and the dragged target is unscaled for an annotative insert so the grip no longer jumps.

Not taken, and the reason this PR is closed rather than merged:

- The repository-wide `cargo fmt` (429 of the 435 files). The tree is deliberately not rustfmt-clean and is not going to be reformatted; a PR that reformats everything cannot be merged.
- The Nix CI rework. It enforces `cargo fmt --check` on every PR, and its flake pins git dependencies by hash — we bump cadcodec and cadkernel several times a week, so it breaks as often as it is fixed (the cadkernel hash here was already stale again). A plain `cargo test --workspace` workflow was added instead, and the unused flake was removed from the repository.

For future changes: one feature per PR, no formatter pass, and it will go in as it is.

PR #1452, opened 2026-09-22
Add audited versioned MCP save workflow

lalomalvi:
## Summary

- add strict explicit DWG/DXF target-version selection to CLI and automation
- add MCP `audit` and `save_verified` operations without expanding the four-tool surface
- validate semantic manifests, layers, blocks, handles, finite bounds, lossy records, and ASCII DXF object graphs
- hash, reopen, version-check, and semantic-compare delivery files
- add repeatable live MCP/GUI and AutoCAD 2025 acceptance harnesses

## Verification

- `cargo test --lib`: 1559 passed, 0 failed, 19 ignored
- MCP schema smoke: passed
- live MCP evaluation: passed (7 tool calls, 1 async task, 13 RPC calls)
- live GUI/MCP acceptance: verified DWG 2000, 2013, 2018 and DXF 2000
- AutoCAD 2025 Core Console AUDIT: 0 errors / 0 fixes for all three DWGs
- cad-core R12 and R2000 fixtures: DWG 2000 round-trip to DXF with ezdxf 0 errors / 0 fixes

## Explicit boundaries

- R14 DWG can be written and reopened, but the observed R14 DWG -> DXF path can orphan object dictionaries. The new audit rejects that output instead of reporting a false green; the verified interoperability lane uses version 2000.
- The live MCP run created and serialized a TEXT entity, but it was not visible in the captured viewport and the command history showed a later text-editor cancellation. This remains a separate visual observation; geometry, manifests, save/reopen, and AutoCAD audits passed.
- AutoCAD Core Console emitted its zero-error result but remained alive after scripted QUIT on the test workstation. The runner isolates Autodesk profile data, records the audit result, and terminates only the process it launched after timeout.
- repository-wide `cargo fmt --check` already reports differences across many untouched files; `git diff --check` passes for this change.

Detailed local verification notes are in `docs/automation/VERIFICATION-2026-09-22.md`.

PR #1453, opened 2026-09-22
Stabilize PLINE viewport integration test stack

lalomalvi:
## Summary
- run the PLINE line-to-arc viewport integration scenario on a dedicated 4 MiB test thread
- keep the exercised application path unchanged
- prevent the Rust test harness default stack from aborting the full suite in debug builds

## Diagnosis
- first bad upstream commit: 652f4d6bbc040dc0b6950086099603dc28fb4185
- the test fails deterministically with the default 2 MiB test-thread stack
- the same binary passes at 3 MiB; this change uses 4 MiB headroom
- failure occurs during the first ViewportLeftRelease after starting PLINE, before ARC mode and before app.view

## Validation
- exact PLINE test: 1 passed
- block-definition/create-block tests: 9 passed
- unexplodable-block regression: 1 passed
- full lib binary: PLINE passes; 1652 passed, 24 ignored, 1 unrelated pre-existing failure
- unrelated failure reproduced on the clean bad-commit binary: psetupin_imports_named_setups_from_a_saved_drawing builds invalid JSON from an unescaped Windows path

`git diff --check` passes. Repository-wide rustfmt check remains red on a pre-existing line in app/automation.rs outside this diff.


lalomalvi:
Combined integration rehearsal against current `upstream/main` passed with #1452, #1454 and #1455 applied too. The full library suite ran **without `RUST_MIN_STACK`**: **1,661 passed, 0 failed, 24 ignored**. This confirms the dedicated PLINE test thread works in the integrated tree.

PR #1454, opened 2026-09-23
Fix PSETUPIN save test on Windows paths

lalomalvi:
## Summary
- serialize the PSETUPIN test save request with `serde_json::json!`
- preserve Windows path separators as valid escaped JSON instead of interpolating `Path::display()` into a raw JSON string

## Root cause
On Windows, the previous request contained a literal path such as `C:\Users\...` inside JSON. The automation parser rejected the first unsupported backslash escape before PSETUPIN could exercise its saved-drawing import path. The product command and file reader were not failing; the test request was malformed.

## Validation
- Original exact test reproduced: `invalid JSON: invalid escape at line 1 column 25`
- Exact PSETUPIN test after fix: 1 passed
- `app::update::page_setup_import::tests::`: 4 passed
- Full library suite: 1,653 passed, 0 failed, 24 ignored

The full suite used `RUST_MIN_STACK=4194304` because upstream `main` does not yet contain the independent PLINE test-stack fix proposed in #1453.


lalomalvi:
Combined integration rehearsal against current `upstream/main` passed with #1452, #1453 and #1455 applied too: **1,661 passed, 0 failed, 24 ignored**. The PSETUPIN Windows-path test passed in that normal full-suite run.

PR #1455, opened 2026-09-23
Keep unit-test plot previews off the user desktop

lalomalvi:
## Summary
- keep normal Plot Preview behavior unchanged in production builds
- make unit-test previews consume their temporary PDF without launching the user's desktop PDF application
- add a regression test that verifies the test preview leaves no temporary file

## Root cause
`app::update::file::plot_paper_tests::sending_a_plot_saves_the_dialog_into_the_layout_when_asked` exercises the real preview pipeline. On native test builds, `open_in_viewer` launched the registered PDF application and retained an `open_cad_studio_preview_<pid>_...pdf` file. Repeated test runs therefore opened blank PDF tabs and accumulated identical files in `%TEMP%`.

## Validation
- new preview-isolation regression: passed
- original plot/page-setup preview test: passed
- preview PDF count around the original test: 5 before, 5 after
- `cargo check --lib` production configuration: passed
- `git diff --check`: passed

This PR is independent from the PSETUPIN Windows JSON fix in #1454 and the PLINE test-stack fix in #1453.


lalomalvi:
Combined integration rehearsal against current `upstream/main` passed with #1452, #1453 and #1454 applied too: **1,661 passed, 0 failed, 24 ignored**. The `open_cad_studio_preview_*.pdf` count was **0 before and 0 after** the full suite, confirming no viewer launch or temp-file leak.

